PR #21, opened 2019-10-08
Use stack to build a statically linked binary of erd

Mogztter:


groovytron:
I think you can merge. Thanks again for this improvement. :+1:

PR #22, opened 2019-10-09
Add note about using docker with sudo

groovytron:
I just added a note about the fact that `sudo` is not needed to build the containers depending on your configuration.

Thanks for your feedback.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -102,6 +102,8 @@ To build all the Docker images, use the following command:
 
  $ sudo make buildDockerImages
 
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -102,6 +102,8 @@ To build all the Docker images, use the following command:
 
  $ sudo make buildDockerImages
 
+`sudo` might not be needed depending on your distribution and `docker` configuration.
```
Mogztter: I think you can add an admonition to draw attention:

```suggestion
NOTE: `sudo` might not be needed depending on your distribution and `docker` configuration.
```

PR #23, opened 2019-10-11
Fix svgbob container build and add build to Travis

groovytron:
# Changes

- Updated `svgbob` from `0.4.1` to `0.4.2`
- `svgbob` installation is now done using `cargo` instead of a `git clone`
- Added container build to Travis Continuous Integration

Thanks in advance for your feedback.

### DIFF
--- a/server/ops/docker/build-static-svgbob
+++ b/server/ops/docker/build-static-svgbob
@@ -1,13 +1,4 @@
 # build static executable binary
 FROM ekidd/rust-musl-builder
 
-RUN git clone https://github.com/ivanceras/svgbob.git /home/rust/src
-
-RUN sudo chown -R rust:rust .
-
-WORKDIR /home/rust/src/svgbob_cli
-
-# version 0.4.1
-RUN git checkout 43fb0364e989d0e9a7656b148c947d47cc769622
-
-RUN cargo build --release
+RUN cargo install --version 0.4.2 svgbob_cli
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: java
 
+services:
+  - docker
+
 jdk: openjdk12
 
 install: true
@@ -7,3 +10,4 @@ install: true
 script:
   - make installLocalDependencies
   - make buildServer
+  - make buildDockerImages
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ buildDockerImages:
 
 tagDockerImages:
 	docker tag kroki-builder-static-erd:latest kroki-builder-static-erd:0.2.0.0
-	docker tag kroki-builder-static-svgbob:latest kroki-builder-static-svgbob:0.4.1
+	docker tag kroki-builder-static-svgbob:latest kroki-builder-static-svgbob:0.4.2
 	docker tag kroki-builder-nomnoml:latest kroki-builder-nomnoml:0.6.1
 
 showExamples:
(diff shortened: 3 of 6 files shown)


server/ops/docker/build-static-svgbob
```diff
@@ -1,13 +1,4 @@
...
-
-RUN sudo chown -R rust:rust .
-
-WORKDIR /home/rust/src/svgbob_cli
-
-# version 0.4.1
-RUN git checkout 43fb0364e989d0e9a7656b148c947d47cc769622
-
-RUN cargo build --release
+RUN cargo install --version 0.4.2 svgbob_cli
```
Mogztter: :+1:

Mogztter: We should probably update the version to 0.4.2 in: 

- https://github.com/yuzutech/kroki/blob/71f661a64c8e7424d3ff59a155494fbb18a0e193/Makefile#L23 
- https://github.com/yuzutech/kroki/blob/71f661a64c8e7424d3ff59a155494fbb18a0e193/server/src/main/java/io/kroki/server/service/HelloHandler.java#L41

And in the documentation: [architecture.adoc#gateway-server](https://github.com/yuzutech/kroki/blob/71f661a64c8e7424d3ff59a155494fbb18a0e193/docs/modules/ROOT/pages/architecture.adoc#gateway-server).

groovytron:
I'll see if I find the time to do it and if you see interest in it but we can refactor the way the containers are built using `docker-compose` and a [build.yaml](https://github.com/groovytron/php-container/blob/master/build.yml). But that's only if you think it simplifies the build.

That approach will delete the numerous `Makefile` used to build each container and centralize it the project's `Makefile` and put the configuration in the `build.yaml`.

Mogztter:
> I'll see if I find the time to do it and if you see interest in it but we can refactor the way the containers are built using docker-compose and a build.yaml. But that's only if you think it simplifies the build.
> That approach will delete the numerous Makefile used to build each container and centralize it the project's Makefile and put the configuration in the build.yaml.

I'm not sure about this one because I still want to be able to (re)build a specific image. I think it makes sense to use a `build.yml` file when you have variations of the "same image" (like in your example with a single `build.yml` that builds an image for PHP 7.1, 7.2 and 7.3) but in this case I'm not so sure... but since I've never used it before I might be wrong 😉

Mogztter:
@groovytron I get the following error on my host :disappointed: 

```console
$ sudo docker build -f server/ops/docker/build-static-svgbob -t kroki-builder-static-svgbob .
Sending build context to Docker daemon  573.9MB
Step 1/2 : FROM ekidd/rust-musl-builder
 ---> 0b63bb1a226d
Step 2/2 : RUN cargo install --version 0.4.2 svgbob_cli
 ---> Running in edb1edb19068
    Updating crates.io index
 Downloading crates ...
  Downloaded svgbob_cli v0.4.2
  Installing svgbob_cli v0.4.2
error: failed to compile `svgbob_cli v0.4.2`, intermediate artifacts can be found at `/tmp/cargo-install0cvFpJ`

Caused by:
  failed to parse lock file at: /home/rust/.cargo/registry/src/github.com-1ecc6299db9ec823/svgbob_cli-0.4.2/Cargo.lock

Caused by:
  invalid serialized PackageId for key `package.dependencies`
The command '/bin/sh -c cargo install --version 0.4.2 svgbob_cli' returned a non-zero code: 101
```

PR #56, opened 2020-02-16
resolves #49 introduces a safe mode

### LINKED ISSUE #49: Support for plantuml include directive on self hosted servers
### About
plantuml's `include` directive is a powerful feature, as illustrated e.g. with [C4](https://github.com/RicardoNiepel/C4-PlantUML) but also in corporate scenarios, allowing to switch diagrams into a common CI, with a single line.

As discussed [here](https://github.com/yuzutech/kroki/issues/33), I think it would be safe to enable (restricted) support for `!include` in self hosted scenarios.

#### include_url
According to [here](http://wiki.plantuml.net/site/preprocessing#including_files) `!include` is also covering the features of `!include_url` meanwhile. We discuss remote and local includes below: 

#### include
In order to safeguard uneducated operators of a self hosted kroki server from unauthorized file reads and cross site attacks  `!include` should  
- only act on a **single preconfigured directory** (e.g. `$PLANTHOME`), if set - and
- accept as parameter **only plain filenames**, without paths.

With those restrictions, I suggest to allow even the configuration of *remote* base urls as `$PLANTHOME` (use case example: Fileservers, with a set of themes).

When there is no such base directory configured, kroki should keep ignoring `!include` lines in diagram bodies (behaviour like now).

#### includesub
This is an `include` variant with the syntax: `!includesub file1.puml!BASIC` where "BASIC" would be only a part of an external definition file.  
I suggest to support it, with the same restrictions as for `!include`.

### Configuration
We are in container age, so I suggest to support as one and only way to configure the presence and location of such a base directory an exported environment variable.

Mogztter:
The safe mode restricts access to files on the file system and on the network.
PlantUML `!include` directive can be enabled using `KROKI_SAFE_MODE=unsafe`.

### DIFF
--- a/docs/modules/setup/pages/manual-install.adoc
+++ b/docs/modules/setup/pages/manual-install.adoc
@@ -38,10 +38,28 @@ java -DKROKI_PORT=1234 -jar kroki-server.jar
 Kroki can be configured using environment variables or Java system properties.
 We've already seen how to change the port using `KROKI_PORT`, let's find out how to configure diagram binary paths.
 
+=== Safe Mode
+
+Kroki provides security levels that restrict access to files on the file system and on the network.
+Each level includes the restrictions enabled in the prior security level.
+
+- `UNSAFE`: disables any security features.
+- `SECURE`: prevents attempts to read files from the file system or from the network.
+
+By default, Kroki is running in `SECURE` mode.
+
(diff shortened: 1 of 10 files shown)


Mogztter:
I think it's good to go!

PR #120, opened 2020-03-17
resolves #117 add Kroki version and SHA1 to hello page

### LINKED ISSUE #117: Show the version of Kroki in the home page
When deploying Kroki from the Docker images, the application shows the version number of the individual components, but not that of Kroki itself.

Mogztter:
kroki-version

resolves #117


akosma:
Thanks a lot! Deployed and it looks great.

PR #130, opened 2020-03-21
Allow to configure the command timeout (5 seconds by default)

Mogztter:


Mogztter:
Should we use a time unit format?

```
KROKI_COMMAND_TIMEOUT=5s
KROKI_COMMAND_TIMEOUT=30000ms
KROKI_COMMAND_TIMEOUT=1m
```

Reference: https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#time-units

PR #149, opened 2020-04-07
Expose supporting containers internally, not publicly

tkfu:
Just a minor thing, but there's no particular reason to publicly expose the supporting container ports (which is what the `ports` directive does). `expose` only exposes the ports to other containers in the same docker network.


Mogztter:
Indeed, since the API of the companion containers is not documented it makes little sense to map the port to the host.

Thanks for the heads up Jon :+1:

Issue #44, opened 2019-12-17
Add BPMN Sketch Miner


hason:
https://design.inf.usi.ch/bpmn-sketch-miner/

Mogztter:
Hello @hason 

It looks very promising!
The syntax is minimal and the diagrams look great 👍 

But is it an open source project? I didn't find much information about the project at https://design.inf.usi.ch/bpmn-sketch-miner/

As far as I understand, this tool is using data mining to understand the described activity/process. But I'm not sure if it's only based on the process you're describing or if the service also train a model with all the processes described. In other words, is it a stateless service or a stateful service.
I guess, we should probably ask Ana Ivanchikj the author of this project?

Mogztter:
I've pinged Ana Ivanchikj on Twitter: https://twitter.com/Mogztter/status/1207041097169620998

ivanchikj:
Hi Guillaume, I just saw your reply on my tweet on BPMN Sketch Miner now, as I got no notification by twitter at the time. Thank you for contacting us. 

Regarding integrating the Sketch Miner in your tool, what type of integration are you interested in? Do you usually get the code of the tools that you integrate or can you use them as a service by calling an API? 

As we are developing the Sketch Miner further we are currently running a survey to see which features would be appreciated and to check user satisfaction (https://design.inf.usi.ch/bpmn-sketch-miner/feedback/feature-vote.html). It does not take more then 15 min, it would be great if you could take a look at it after having played around with the tool. Not sure when is the last time you checked out our tool, but we have added some tutorials and examples to the same to get you started.

Mogztter:
Hello @ivanchikj

> Regarding integrating the Sketch Miner in your tool, what type of integration are you interested in? Do you usually get the code of the tools that you integrate or can you use them as a service by calling an API?

I usually get the code as Kroki is exclusively using open source tools.
I don't really mind which language is used as long as it's relatively easy to produce an image from a text description. I guess I could also call an API but I would rather not rely on a third party service as Kroki can be installed on a local environment with a restricted network.

> As we are developing the Sketch Miner further we are currently running a survey to see which features would be appreciated and to check user satisfaction (https://design.inf.usi.ch/bpmn-sketch-miner/feedback/feature-vote.html). It does not take more then 15 min, it would be great if you could take a look at it after having played around with the tool. 

I'm not familiar with all the concepts in the survey but I will do my best to fill it.

[...]

PR #178, opened 2020-04-24
Add BPMN support

nekator:
add support to render BPMN diagram via bpmn-js
GH-44


Mogztter:
Thank you very much @nekator 
It looks great! :+1: 

I will do a more in depth review this week.

nekator:
You are welcome.  Looking forward for your feedback. 😀

server/src/main/java/io/kroki/server/service/Bpmn.java
```diff
@@ -0,0 +1,53 @@
...
+    this.client = WebClient.create(vertx);
+    this.sourceDecoder = new SourceDecoder() {
+      @Override
+      public String decode(String encoded) throws DecodeException {
+        return DiagramSource.decode(encoded);
+      }
+    };
+    this.host = config.getString("KROKI_BPMN_HOST", "127.0.0.1");
+    this.port = config.getInteger("KROKI_BPMN_PORT", 8003);
+    this.diagramResponse = new DiagramResponse(new Caching("8.4.8"));
```
Mogztter, line 36: The version number should be `6.4.2` right?

Mogztter:
Sorry I didn't have time to give it a try yet but since it's a new container we will need to update the following pages:

- https://github.com/yuzutech/kroki/blob/master/docs/modules/setup/pages/install.adoc#images
- https://github.com/yuzutech/kroki/blob/master/docs/modules/ROOT/pages/architecture.adoc

server/src/main/java/io/kroki/server/service/Bpmn.java
nekator: Yes. It is a copy and paste error. I will change it to the correct version.

nekator:
I'll take a look into the docs and update the relevant parts today or tomorrow.

server/src/main/java/io/kroki/server/service/Bpmn.java
nekator: What does version mean in this context? If it is diagramm version then it would be `2.0`.
If it means the version of the underlaying libary it would be `6.4.2`

nekator:
I updated the docs for install and architecture. I also changed the caching version to the diagram version.

bpmn/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "scripts": {
+    "prestart": "cp node_modules/bpmn-js/dist/bpmn-viewer.production.min.js assets/bpmn-viewer.production.min.js",
+    "start": "node src/index.js",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "Heiko Does",
+  "license": "MIT",
+  "dependencies": {
+    "bpmn-js": "^6.4.2",
```
Mogztter, line 16: Could you please update to 6.5.1?

docs/modules/ROOT/pages/architecture.adoc
```diff
@@ -104,3 +107,14 @@ The Docker image contains the following diagrams libraries out-of-the-box:
...
+
+=== BPMN
+
+The Docker image contains the following diagrams libraries out-of-the-box:
+
+[options="header",cols="1,1m"]
+|===
+|Diagram Type | Version
+|https://bpmn.io/toolkit/bpmn-js/[BPMN]
+|2.0
```
Mogztter, line 119: I think we should use the actual version of the library: 6.5.1

server/src/main/java/io/kroki/server/service/Bpmn.java
```diff
@@ -0,0 +1,53 @@
...
+    this.client = WebClient.create(vertx);
+    this.sourceDecoder = new SourceDecoder() {
+      @Override
+      public String decode(String encoded) throws DecodeException {
+        return DiagramSource.decode(encoded);
+      }
+    };
+    this.host = config.getString("KROKI_BPMN_HOST", "127.0.0.1");
+    this.port = config.getInteger("KROKI_BPMN_PORT", 8003);
+    this.diagramResponse = new DiagramResponse(new Caching("2.0"));
```
Mogztter, line 36: Can you please use 6.5.1 here?

bpmn/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+FROM alpine:3.11.6
+
+COPY --from=builder /app/app.bin /node/bin/bpmn
+COPY assets /app/assets
+ENV KROKI_BPMN_PAGE_URL=file:///app/assets/index.html
+
+RUN apk add --no-cache --update chromium
+ENV PUPPETEER_EXECUTABLE_PATH=/usr/lib/chromium/chrome
+
+EXPOSE 8002
```
Mogztter, line 35: Should be 8003 right?

Mogztter:
Overall it's looking very good, thanks @nekator 
I left a few comments but I can take care of them if you want.

I wonder if we should remove the XML header from the SVG. Currently, `bpmn-js` will generate the following SVG:

```xml
<?xml version="1.0" encoding="utf-8"?>
<!-- created with bpmn-js / http://bpmn.io -->
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="856" height="587" viewBox="150 75 856 587" version="1.1">
<!-- ... -->
```

Maybe it can be configured...

nekator:
I will and the relevant changes tomorrow. 

Regarding XML header i would suggest to keep it.

Maybe we should make it configurable to support other output formats because bpmn-js support multiple formats.

nekator:
Resolved everything and squashed it to a single feature commit.

Mogztter:
> Regarding XML header i would suggest to keep it.

I don't have a strong opinion, so we will keep it for now.

> Maybe we should make it configurable to support other output formats because bpmn-js supports multiple formats.

Oh nice, I didn't know!
@nekator I guess that since they are using a `<canvas>` we can export it to png using [`toDataURL`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL). Is that what you had in mind?
But as far as I understand, it's only possible to save as SVG using the `Viewer` API (ie. [`BaseViewer.prototype.saveSVG`](https://github.com/bpmn-io/bpmn-js/blob/f3edfd0f0d3c4785bec6e2408a3bad2e86b31769/lib/BaseViewer.js#L425)) right? 

I think we can add additional output formats in follow-up pull requests. Wdyt?

nekator:
Adding additional formats would be another PR IMO. For now it is good to have BPMN diagram to svg. I will take a look into other formats maybe this weekend. 

I'm glad you like the code and that you took the time to review it properly.  Thank you 😊

PR #187, opened 2020-04-28
Use a simpler setup for Docker

Mogztter:
`actions-setup-docker` seems unstable/heavyweight and had failed quite a lot recently so trying a new approach taken from the "official" Docker GitHub Actions repository: https://github.com/docker/github-actions


Mogztter:
What do you think about this approach @groovytron? The good part is that we are down from more than 1 minute to a few seconds but maybe we are missing some steps?

groovytron:
LGTM @Mogztter. I think it's a good improvement. Well done! :+1:

Mogztter:
Thanks for your feedback,merging!

PR #207, opened 2020-05-05
Allow configuration of PlantUML includes

stephenc:
We generate images as part of our build chain and so we trust the URLs and other libraries that we want to use as includes. The current includes blocking is heavy handed. This should allow import path includes as well as enabling a whitelist of regexes for anyone who wants to allow including only a subset of URLs that they actually trust.

Note: I would advise repeating my analysis of PlantUML's `ImportedFiles` to verify that for regular file imports it only permits the inclusion of files that are actually on the include path (my analysis showed that it is a bit too strict as you cannot have `!include foo/bar.puml` because the `getSystemFolder()` will then return `$DIR_FROM_SEARCH_PATH/foo` which is not explicitly one of the files on the search path... but that seems sufficiently strict from my PoV to allow any paths that do not contain `..` style navigation and are not URLs

### DIFF
--- a/server/README.adoc
+++ b/server/README.adoc
@@ -55,7 +55,7 @@ When one of the above library is requested, Kroki server will delegate the call.
 
 ==== PlantUML include files
 
-As PlantUML syntax includes pre-processor directives that have the potential to access external resources, serving live content from Kroki server is only recommended when using the dfeault settings of `KROKI_SAFE_MODE=SECURE` and `KROKI_PLANTUML_ALLOW_INCLUDE=false`.
+As PlantUML syntax includes pre-processor directives that have the potential to access external resources, serving live content from Kroki server is only recommended when using the default settings of `KROKI_SAFE_MODE=SECURE` and `KROKI_PLANTUML_ALLOW_INCLUDE=false`.
 
 If you are using Kroki server for offline image generation as part of your build tool chain, you may be able to change these defaults although it is not recommended.
 
@@ -79,5 +79,4 @@ If you don't want to use the Docker image, you will need to install all the depe
 * GraphViz: https://graphviz.gitlab.io/download/
 * Erd: https://github.com/BurntSushi/erd
 * Svgbob: https://github.com/ivanceras/svgbob
-
 
--- a/server/README.adoc
+++ b/server/README.adoc
@@ -28,6 +28,10 @@ KROKI_DOT_BIN_PATH:: Path to the `dot` binary (provided by GraphViz)
 KROKI_NOMNOML_BIN_PATH:: Path to the `nomnoml` binary
 KROKI_BLOCKDIAG_HOST:: The host of the blockdiag server
 KROKI_BLOCKDIAG_PORT:: The port of the blockdiag server
+KROKI_SAFE_MODE:: Either `SECURE` (default) or `UNSAFE`. Determines whether to sanitize requests for formats that allow including external content
+KROKI_PLANTUML_INCLUDE_PATH:: The include path to set for PlantUML.
+KROKI_PLANTUML_INCLUDE_WHITELIST:: The name of a file that consists of a list or regexes for valid includes.
+KROKI_PLANTUML_ALLOW_INCLUDE:: Either `false` (default) or `true`. Determines if PlantUML will fetch `!include` directives that reference external URLs
 
 === Image
 
@@ -48,6 +52,24 @@ To keep the Docker image as small as possible, the following libraries are provi
  * RackDiag
 
 When one of the above library is requested, Kroki server will delegate the call.
+
+==== PlantUML include files
+
+As PlantUML syntax includes pre-processor directives that have the potential to access external resources, serving live content from Kroki server is only recommended when using the dfeault settings of `KROKI_SAFE_MODE=SECURE` and `KROKI_PLANTUML_ALLOW_INCLUDE=false`.
+
+If you are using Kroki server for offline image generation as part of your build tool chain, you may be able to change these defaults although it is not recommended.
+
+The recommended pattern for handling includes is to either extend the docker image or bind mount a directory containing the files you wish to include and then set the `KROKI_PLANTUML_INCLUDE_PATH` environment variable.
+
+For example:
+
+[source,bash]
+----
+$ mkdir example-lib
+$ printf "@startuml\nBob->Alice\n@enduml\n" > example-lib/bob.puml
+$ sudo docker run --publish=8000:8000 -v $(pwd)/example-lib:/example-lib -e KROKI_PLANTUML_INCLUDE_PATH=/example-lib yuzutech/kroki:latest
+$ curl http://localhost:8000/plantuml/svstartuml\n!include bob.puml\nAlice -> Bob\n@enduml' | python -c "import sys; import base64; import zlib; print(base64.urlsafe_b64encode(zlib.compress(sys.stdin.read(), 9)))")
+----
 
 == Manual install
 
--- a/server/README.adoc
+++ b/server/README.adoc
@@ -28,9 +28,27 @@ KROKI_DOT_BIN_PATH:: Path to the `dot` binary (provided by GraphViz)
 KROKI_NOMNOML_BIN_PATH:: Path to the `nomnoml` binary
 KROKI_BLOCKDIAG_HOST:: The host of the blockdiag server
 KROKI_BLOCKDIAG_PORT:: The port of the blockdiag server
-KROKI_SAFE_MODE:: Either `SECURE` (default) or `UNSAFE`. Determines whether to sanitize requests for formats that allow including external content
+KROKI_SAFE_MODE:: One of:
++
+--
+`SECURE`:: (default) Paranoid sanitization of requests before passing to image server.
+`SAFE`:: Assume the image servers secure mode request sanitization is sufficient
+`UNSAFE`:: Run the image servers without request sanitization.
+--
++
+[NOTE]
+--
+Some of the image servers allow referencing external entities by URL or accessing resources from the filesystem.
+
+For example PlantUML allows the `!import` directive to pull fragments from the filesystem or a remote URL or the standard library.
+
+It is the responsibility of the upstream codebases to ensure that they can be safely used without risk.
+Because Kroki does not perform code review of these servies, our default setting is to be paranoid and block imports unless known safe.
+We encourage anyone running their own Kroki server to review the services security settings and select the security mode appropriate for their use case.
+--
 KROKI_PLANTUML_INCLUDE_PATH:: The include path to set for PlantUML.
 KROKI_PLANTUML_INCLUDE_WHITELIST:: The name of a file that consists of a list or regexes for valid includes.
+KROKI_PLANTUML_INCLUDE_WHITELIST_0, KROKI_PLANTUML_INCLUDE_WHITELIST_1, ... KROKI_PLANTUML_INCLUDE_WHITELIST___N__:: One regex to add to the include whitelist per environment variable. Search will stop at the first empty or undefined integer number.
 KROKI_PLANTUML_ALLOW_INCLUDE:: Either `false` (default) or `true`. Determines if PlantUML will fetch `!include` directives that reference external URLs
 
 === Image
--- a/server/README.adoc
+++ b/server/README.adoc
@@ -85,7 +85,7 @@ For example:
 ----
 $ mkdir example-lib
 $ printf "@startuml\nBob->Alice\n@enduml\n" > example-lib/bob.puml
-$ sudo docker run --publish=8000:8000 -v $(pwd)/example-lib:/example-lib -e KROKI_PLANTUML_INCLUDE_PATH=/example-lib yuzutech/kroki:latest
+$ sudo docker run --publish=8000:8000 -v $(pwd)/example-lib:/example-lib -e KROKI_SAFE_MODE=SAFE -e KROKI_PLANTUML_INCLUDE_PATH=/example-lib yuzutech/kroki:latest
 $ curl http://localhost:8000/plantuml/svg/$(printf '@startuml\n!include bob.puml\nAlice -> Bob\n@enduml' | python -c "import sys; import base64; import zlib; print(base64.urlsafe_b64encode(zlib.compress(sys.stdin.read(), 9)))")
 ----
 
@@ -97,4 +97,3 @@ If you don't want to use the Docker image, you will need to install all the depe
 * GraphViz: https://graphviz.gitlab.io/download/
 * Erd: https://github.com/BurntSushi/erd
 * Svgbob: https://github.com/ivanceras/svgbob
-
--- a/server/README.adoc
+++ b/server/README.adoc
@@ -31,23 +31,23 @@ KROKI_BLOCKDIAG_PORT:: The port of the blockdiag server
 KROKI_SAFE_MODE:: One of:
 +
 --
-`SECURE`:: (default) Paranoid sanitization of requests before passing to image server.
-`SAFE`:: Assume the image servers secure mode request sanitization is sufficient
-`UNSAFE`:: Run the image servers without request sanitization.
+`SECURE`:: (default) Paranoid sanitization of requests before passing to the diagram library.
+`SAFE`:: Assume the diagram libraries secure mode request sanitization is sufficient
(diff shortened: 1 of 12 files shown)


Mogztter:
Wow thanks for this great contribution!
This is a popular request and we will probably extend it (in the future) to Vega and Vega-Lite which are using a similar feature to include data from URL.

Overall I like the general concept but I would like to keep the `SECURE` safe mode as is and introduce a new level `SAFE` where the include path and the include whitelist will be taken into consideration.

Basically, `SECURE` level does not trust the underlying library.
It's not really related to this pull request but, just for reference, on the public instance of Kroki I would like to deny access to all/most filesystem operations using Docker (ie. remove Linux kernel capabilities).

server/README.adoc
```diff
@@ -49,6 +53,24 @@ To keep the Docker image as small as possible, the following libraries are provi
 
 When one of the above library is requested, Kroki server will delegate the call.
 
+==== PlantUML include files
+
+As PlantUML syntax includes pre-processor directives that have the potential to access external resources, serving live content from Kroki server is only recommended when using the dfeault settings of `KROKI_SAFE_MODE=SECURE` and `KROKI_PLANTUML_ALLOW_INCLUDE=false`.
```
Mogztter, line 58:
```suggestion
As PlantUML syntax includes pre-processor directives that have the potential to access external resources, serving live content from Kroki server is only recommended when using the default settings of `KROKI_SAFE_MODE=SECURE` and `KROKI_PLANTUML_ALLOW_INCLUDE=false`.
```

server/src/main/java/io/kroki/server/service/Plantuml.java
```diff
@@ -24,25 +24,52 @@
...
+   *   <li>{@code !include /absolute/file/path} includes a file from the filesystem</li>
+   *   <li>{@code !include search-path-file} includes a file on the "plantuml.include.path"</li>
+   *   <li>{@code !include search-path-file!2} includes the 3rd document in a file on the "plantuml.include.path"</li>
+   *   <li>{@code !includesub search-path-file!SUBID} includes the SUBID subpart of a file on the "plantuml.include.path"</li>
+   *   <li>{@code !includeurl http://some/url} deprecated include of an external resource by URL</li>
+   * </ul>
+   *
+   * @see <a href="https://plantuml.com/preprocessing">PlantUML Preprocessing</a>
+   */
+  private static final Pattern INCLUDE_RX = Pattern.compile("^\\s*!include(?:url|sub)?\\s+(.*)(?:!.*)?(?:#.*)?");
```
Mogztter, line 66: Much needed documentation, thanks :+1: 
What is the last part of the regular expression with the `#` symbol? Is that when a URL contains a fragment?

server/README.adoc
```diff
@@ -28,6 +28,10 @@ KROKI_DOT_BIN_PATH:: Path to the `dot` binary (provided by GraphViz)
 KROKI_NOMNOML_BIN_PATH:: Path to the `nomnoml` binary
 KROKI_BLOCKDIAG_HOST:: The host of the blockdiag server
 KROKI_BLOCKDIAG_PORT:: The port of the blockdiag server
+KROKI_SAFE_MODE:: Either `SECURE` (default) or `UNSAFE`. Determines whether to sanitize requests for formats that allow including external content
+KROKI_PLANTUML_INCLUDE_PATH:: The include path to set for PlantUML.
+KROKI_PLANTUML_INCLUDE_WHITELIST:: The name of a file that consists of a list or regexes for valid includes.
```
Mogztter, line 33: I guess that passing a list of regexes as environment variable is not practical but using a file means that you need to extend the docker image or bind mount a directory with this file right?
Is there an alternative solution?

server/src/main/java/io/kroki/server/service/Plantuml.java
```diff
@@ -65,8 +92,26 @@ public String decode(String encoded) throws DecodeException {
...
+    }
+  }
+
+  private static List<Pattern> parseIncludeWhitelist(String filename) {
+    if (filename != null && new File(filename).isFile()) {
+      try {
+        return Files.lines(Paths.get(filename)).filter(s -> !s.isEmpty()).map(Pattern::compile)
+          .collect(Collectors.toList());
+      } catch (IOException e) {
+        // ignore
```
Mogztter, line 111: I think we should log the error with the context, so the user knows what is going on:

```
logger.warn("Unable to read the PlantUML whitelist file: xyz. The include whitelist will be empty.", e)
```

server/src/main/java/io/kroki/server/service/Plantuml.java
```diff
@@ -65,8 +92,26 @@ public String decode(String encoded) throws DecodeException {
...
+    OptionFlags.ALLOW_INCLUDE = config.getBoolean("KROKI_PLANTUML_ALLOW_INCLUDE", false);
+    String plantUmlIncludePath = config.getString("KROKI_PLANTUML_INCLUDE_PATH");
+    if (plantUmlIncludePath != null) {
+      // NOTE: relying on classloading sequencing to ensure this gets set before ImportedFiles class is loaded by JVM
+      System.setProperty("plantuml.include.path", plantUmlIncludePath);
+    }
+  }
+
+  private static List<Pattern> parseIncludeWhitelist(String filename) {
+    if (filename != null && new File(filename).isFile()) {
```
Mogztter, line 106: If the filename is not a file or is not readable I also think we should add a log.

server/src/main/java/io/kroki/server/service/Plantuml.java
```diff
@@ -177,6 +226,43 @@ private static void ignoreInclude(String line, StringBuilder sb) {
...
+        //   a fixed set guarded by `/stdlib/${name}-abx.repx` resources on the classpath, in any case we
+        //   ignore stdlib import validation here
+        // * Absolute imports could refer to any file, so we exclude those, though FileWithSuffix#fileOk()
+        //   should restrict accessible files to only those "approved"
+        // * We only auto-permit "search path" includes that do not attempt search path escaping with `..`
+        // * The import or importsub that we auto-permit will result in a call to ImportedFiles.getFile(...)
+        // * That will iterate the "plantuml.include.path" (note PlantUML is not well designed for embedding
+        //   on this property as the ImportedFiles#INCLUDE_PATH is initialized once on classloading)
+        // * At this point, if the INCLUDE_PATH included say `/example` and the include was `foo/bar.puml`
+        //   and there is a file `/example/foo/bar/puml` then that would stop the search with an `AFileRegular`
```
Mogztter, line 253: you meant `/example/foo/bar.puml`?

server/src/main/java/io/kroki/server/service/Plantuml.java
```diff
@@ -177,6 +226,43 @@ private static void ignoreInclude(String line, StringBuilder sb) {
...
+        //   instance. HOWEVER, the return value of ImportedFiles#getFile is guarded by ImportedFiles#isAllowed
+        // * ImportedFiles#isAllowed will only permit AFile instances with the AFile#getSystemFolder() being
+        //   contained in ImportedFiles#INCLUDE_PATH so as AFileRegular#getSystemFolder() always returned the
+        //   parent folder of the file and include of `foo/bar.puml` found on an include path of `/example`
+        //   resolving to `/example/foo/bar.puml` will have a system folder of `/example/foo` which is not
+        //   one of the folders whitelisted in the include path and thus the include will be resolved as not-found
+        //
+        // Effectively only imports that are immediate children of the folders listed in "plantuml.include.path"
+        // are eligible for inlcude (unless OptionFlags.ALLOW_INCLUDE has been set to true)
+        sb.append(line).append("\n");
```
Mogztter, line 263: I really appreciate the time and effort you put into this clear and precise analysis of the PlantUML include checking algorithm :100: 

If I understand correctly, even if we do allow path walking, the PlantUML include algorithm should not resolve the include if the path is outside of the include path right? So if my include path is `/example` and my include is `../etc/passwd`, the system folder of the file will be `/etc` and `/etc` is not part of the include path `/example` so the include will be resolved as not-found. Am I correct?
I'm not saying that we should remove the path walking checks (as we want to disallow them as soon as possible) but I want to make sure that I understand how it's working.

server/src/main/java/io/kroki/server/service/Plantuml.java
stephenc: that's a comment in plantuml

server/README.adoc
stephenc: you could have one regex per env-var and chase up env vars with indexes, e.g. `KROKI_PLANTUML_INCLUDE_WHITELIST_0`, `KROKI_PLANTUML_INCLUDE_WHITELIST_1`, until you stop finding `KROKI_PLANTUML_INCLUDE_WHITELIST_n` for some integer n

server/src/main/java/io/kroki/server/service/Plantuml.java
stephenc: correct, but please verify my claim rather than take it at face value. I believe it to be correct, but you are the maintainer, so "trust but verify" especially as this as security implications

server/README.adoc
```diff
@@ -49,6 +71,24 @@ To keep the Docker image as small as possible, the following libraries are provi
...
+
+The recommended pattern for handling includes is to either extend the docker image or bind mount a directory containing the files you wish to include and then set the `KROKI_PLANTUML_INCLUDE_PATH` environment variable.
+
+For example:
+
+[source,bash]
+----
+$ mkdir example-lib
+$ printf "@startuml\nBob->Alice\n@enduml\n" > example-lib/bob.puml
+$ sudo docker run --publish=8000:8000 -v $(pwd)/example-lib:/example-lib -e KROKI_PLANTUML_INCLUDE_PATH=/example-lib yuzutech/kroki:latest
```
stephenc, line 88:
```suggestion
$ sudo docker run --publish=8000:8000 -v $(pwd)/example-lib:/example-lib -e KROKI_SAFE_MODE=SAFE -e KROKI_PLANTUML_INCLUDE_PATH=/example-lib yuzutech/kroki:latest
```

stephenc:
Anything else?

Mogztter:
> Anything else?

No it's looking really good! :+1: 
I will probably do a thorough review/testing session this weekend and I will also take time to read the source code of PlantUML. I will let you know :wink:

Mogztter:
@stephenc I'm looking at it right now, I will probably do some small changes and commit them.

server/src/main/java/io/kroki/server/service/Plantuml.java
```diff
@@ -153,30 +229,73 @@ static String withDelimiter(String source) {
...
+      } else if (includeWhitelist.stream().anyMatch(p -> p.matcher(include).matches())) {
```
Mogztter, line 297: Not sure which parameter should have the priority between the safe mode and the include whitelist options. I am somewhat leaning toward safe mode. In other words, I think that if safe mode SECURE is defined we should ignore whitelist. Arguably it's a bit restrictive but if it's clearly stated in the documentation that's OK.
Plus, if we do the opposite you cannot have this behavior.

Mogztter: I went ahead and now safe mode SECURE has the priority over the whitelist.

server/README.adoc
```diff
@@ -28,6 +28,28 @@ KROKI_DOT_BIN_PATH:: Path to the `dot` binary (provided by GraphViz)
 KROKI_NOMNOML_BIN_PATH:: Path to the `nomnoml` binary
 KROKI_BLOCKDIAG_HOST:: The host of the blockdiag server
 KROKI_BLOCKDIAG_PORT:: The port of the blockdiag server
+KROKI_SAFE_MODE:: One of:
++
+--
+`SECURE`:: (default) Paranoid sanitization of requests before passing to image server.
```
Mogztter, line 34: I have used "diagram library" instead of "image server" in the documentation. I'm not a native speaker so maybe "image server" is more appropriated but in this case, we should replace every occurrences of "diagram library" otherwise it might be confusing.

Maybe we can stick with "diagram library" in this pull request and create another one to change the terminology if needed? What do you think?

server/README.adoc
```diff
@@ -28,6 +28,28 @@ KROKI_DOT_BIN_PATH:: Path to the `dot` binary (provided by GraphViz)
...
+--
++
+[NOTE]
+--
+Some of the image servers allow referencing external entities by URL or accessing resources from the filesystem.
+
+For example PlantUML allows the `!import` directive to pull fragments from the filesystem or a remote URL or the standard library.
+
+It is the responsibility of the upstream codebases to ensure that they can be safely used without risk.
+Because Kroki does not perform code review of these servies, our default setting is to be paranoid and block imports unless known safe.
```
Mogztter, line 46:
```suggestion
Because Kroki does not perform code review of these services, our default setting is to be paranoid and block imports unless known safe.
```

server/README.adoc
```diff
@@ -28,6 +28,28 @@ KROKI_DOT_BIN_PATH:: Path to the `dot` binary (provided by GraphViz)
...
+We encourage anyone running their own Kroki server to review the services security settings and select the security mode appropriate for their use case.
```
Mogztter, line 47: Not sure about this one:

```suggestion
We encourage anyone running their own Kroki server to review the services security settings and select the appropriated security mode for their use case.
```

server/README.adoc
```diff
@@ -28,6 +28,28 @@ KROKI_DOT_BIN_PATH:: Path to the `dot` binary (provided by GraphViz)
...
+--
+KROKI_PLANTUML_INCLUDE_PATH:: The include path to set for PlantUML.
+KROKI_PLANTUML_INCLUDE_WHITELIST:: The name of a file that consists of a list or regexes for valid includes.
```
Mogztter, line 50:
```suggestion
KROKI_PLANTUML_INCLUDE_WHITELIST:: The name of a file that consists of a list of Java regular expressions for valid includes.
```

server/README.adoc
stephenc: As a native English speaker, don’t make this change

Mogztter: Alright :baguette_bread: :fr:

server/README.adoc
Mogztter: So stick with "diagram library"?

Mogztter:
```suggestion
`SECURE`:: (default) Paranoid sanitization of requests before passing to the diagram library.
```

server/README.adoc
```diff
@@ -28,6 +28,28 @@ KROKI_DOT_BIN_PATH:: Path to the `dot` binary (provided by GraphViz)
 KROKI_NOMNOML_BIN_PATH:: Path to the `nomnoml` binary
 KROKI_BLOCKDIAG_HOST:: The host of the blockdiag server
 KROKI_BLOCKDIAG_PORT:: The port of the blockdiag server
+KROKI_SAFE_MODE:: One of:
++
+--
+`SECURE`:: (default) Paranoid sanitization of requests before passing to image server.
+`SAFE`:: Assume the image servers secure mode request sanitization is sufficient
```
Mogztter, line 35:
```suggestion
`SAFE`:: Assume the diagram libraries secure mode request sanitization is sufficient
```

server/README.adoc
```diff
@@ -28,6 +28,28 @@ KROKI_DOT_BIN_PATH:: Path to the `dot` binary (provided by GraphViz)
...
+`UNSAFE`:: Run the image servers without request sanitization.
```
Mogztter, line 36:
```suggestion
`UNSAFE`:: Run the diagram libraries without request sanitization.
```

server/README.adoc
```diff
@@ -28,6 +28,28 @@ KROKI_DOT_BIN_PATH:: Path to the `dot` binary (provided by GraphViz)
...
+--
++
+[NOTE]
+--
+Some of the image servers allow referencing external entities by URL or accessing resources from the filesystem.
```
Mogztter, line 41:
```suggestion
Some of the diagram libraries allow referencing external entities by URL or accessing resources from the filesystem.
```

Mogztter:
@stephenc I think it's ready to go. I made a few changes, I will let you review them before merging.

Mogztter:
I got a thumb up from @stephenc so let's do it!
A big thank you Stephen for your work :medal_sports:

PR #214, opened 2020-05-10
Revert "Switch to Non-Maven Jar plugin to vendor jars"

Mogztter:
IntelliJ cannot properly resolve the dependencies when using the `non-maven-jar-maven-plugin`, so reverting for now.
Please note that this is a bug in IDEA (as it's working fine in Maven). Probable root cause: https://youtrack.jetbrains.com/issue/IDEA-175538

> I am sorry, IntelliJ seems to have taken a step back in how they parse Maven project models. I may need to rework some hacks into the non-maven jar. For now just revert this commit and I'll re-submit the PR once I find some way to get IDEA working again...

Reverts yuzutech/kroki#206


stephenc:
https://youtrack.jetbrains.com/issue/IDEA-228084 when fixed would at least provide for using a profile that is disabled in intellij pending a deeper fix of their classpath artifact resolution code

stephenc:
https://youtrack.jetbrains.com/issue/IDEA-228084 is the ultimate root cause requiring this revert

PR #279, opened 2020-06-21
kroki-cli doc: Link to releases page

ferdnyc:
Instead of linking to a specific version (and falling out of date), just link to the releases listing, which will show latest on top.


Mogztter:
Indeed that's better, one less thing to update and eventually forget!
Thanks for your contribution 👍

PR #329, opened 2020-07-26
resolves #297 add an health check API

### LINKED ISSUE #297: Add an API that returns the diagram librairies versions and the server version
Currently, we only have an HTML page https://a.kroki.io but it could be useful to have an endpoint if we want to use that information programmatically.

For instance, we could locally cache images depending on the diagram library version:

```
plantuml-1-2020-10/
  54cae01691d18960adccb024b0e5b01b.png

plantuml-1-2020-15/
  54cae01691d18960adccb024b0e5b01b.png
```

A few ideas:

- `/about`
- `/version`
- `/versions`

Alternatively, we could also use a "status" route to check that the server is alive and we could use this opportunity to return server infos:

- `/status`
- `/_status`

Mogztter:
This API returns a status and a list of versions.

resolves #297


llaville:
Hello @Mogztter
What is the status of this PR ?
Friendly ping

Mogztter:
I had second thoughts about the API versioning :sweat_smile: 
I think I'm going to add two aliases:

* `/healthz` to be "compatible" with K8s ecosystem
* `/v1/health` safe option for tooling

Thoughts?

llaville:
> K8s ecosystem
Sorry but I cannot said anything, because I don't know Kubernetes. My skill is only on Docker  

> safe option for tooling
What do you means by this alias ?

Mogztter:
> What do you means by this alias?

Basically `/health` is an alias on the latest version of the health API: `/health` -> `/v1/health`.

If at one point in the future we decide to introduce a breaking change in the health API, we can add a new endpoint `/v2/health`. The `/health` alias will then be updated to resolve to `/v2/health`.

As a consequence, the safest option, if you rely on this API, is to use an explicit version: `/v1/health`.
But maybe it's far fetched...?

llaville:
**Versioning our API is always a great debate !**

In this occasion, I would like to share a link (if you don't know yet) about [API design](https://github.com/dwyl/learn-api-design) !
And the part that we want to focus now [Versioning Your API](https://github.com/dwyl/learn-api-design#versioning-your-api). 

I suggest you to read the first article link called [Your API versioning is wrong](https://www.troyhunt.com/your-api-versioning-is-wrong-which-is/) .
I'm agree with author that we let choose "clients" how to consume API, but not forced a way unless there is a good fallback. 
I've in mind the “tabs versus spaces” debate and the solution for all developers : [EditorConfig](https://editorconfig.org/).

Mogztter:
Nice reading, thanks for sharing.

I think for now the alias plus the URL versioning give enough options for the consumer of the API. If needed we could use a custom request header or a custom content-type depending on how the API will evolve (or not).

Mogztter:
@llaville Thanks again for your input on this issue, greatly appreciated :hugs:

llaville:
@Mogztter Glad to shared

PR #569, opened 2021-01-26
Fix dependencies and Excalidraw integration in Docker Compose example

marcelstoer:


docs/modules/setup/examples/kroki-docker-compose.yml
```diff
@@ -1,11 +1,17 @@
 version: "3"
 services:
-  core:
+  kroki:
     image: yuzutech/kroki
+    depends_on:
+      - blockdiag
+      - mermaid
+      - bpmn
+      - excalidraw
```
Mogztter, line 9: As far as I know it's not strictly required right? Did you have any issue while running `docker-compose up -d` without this option?

The Docker compose documentations states:

> `depends_on` does not wait for db and redis to be “ready” before starting web - only until they have been started. If you need to wait for a service to be ready, see Controlling startup order for more on this problem and strategies for solving it.

So, `blockdiag`, `mermaid`, `bpmn` and `excalidraw` will start first but `kroki` will still be started before they are ready (i.e, before the companions containers can receive requests).
So I'm not sure that it's worth to add this option without using the [`command` option](https://docs.docker.com/compose/startup-order/).

docs/modules/setup/examples/kroki-docker-compose.yml
```diff
@@ -1,11 +1,17 @@
...
     environment:
       - KROKI_BLOCKDIAG_HOST=blockdiag
       - KROKI_MERMAID_HOST=mermaid
       - KROKI_BPMN_HOST=bpmn
+      - KROKI_EXCALIDRAW_HOST=excalidraw
```
Mogztter, line 14: :+1:

docs/modules/setup/examples/kroki-docker-compose.yml
marcelstoer: No issues, they all start sufficiently fast. However, we use `depends_on` to clearly express dependencies between containers to anyone looking at the compose file. As such its primary purpose is documentation.

Mogztter: Technically, the gateway container `kroki` can start without the companion containers but as you said it can emphasis the fact that the `kroki` container relies on a list of companion containers to provide additional diagrams.

Let's keep it for now :wink:

Mogztter:
Thanks @marcelstoer!

Mogztter:
It's live: https://docs.kroki.io/kroki/setup/install/ :tada:

PR #579, opened 2021-02-02
resolves #386 install CJK font

### LINKED ISSUE #386: Japanese text result is incorrect for png format
original diagram

```
digraph D {
   b1 [label = "第一次川中島の戦い\n1553年"];
   b2 [label = "第二次川中島の戦い\n1555年"];
   b4 [label = "第三次川中島の戦い\n1557年"];
   b1 -> b2 -> b4;
}
```

compare PNG and SVG

https://kroki.io/graphviz/svg/eNpLyUwvSizIUHBRqOZSUFBIMlSIzklMSs1RsFVQer5mzZMdDc_WLHy6fe6THWufbt72uHHds45ljxtbYvIMTU2Nn-7cohRrDdZnhKZvVw9ufaZI-kzQ7evErc8cSZ-hgq4dyFYQaWLNVQsALpNctA==

https://kroki.io/graphviz/png/eNpLyUwvSizIUHBRqOZSUFBIMlSIzklMSs1RsFVQer5mzZMdDc_WLHy6fe6THWufbt72uHHds45ljxtbYvIMTU2Nn-7cohRrDdZnhKZvVw9ufaZI-kzQ7evErc8cSZ-hgq4dyFYQaWLNVQsALpNctA==

Mogztter:
With this change, CJK characters are now correctly rendered when converting to PNG:

out

For reference, here's the result using http://www.plantuml.com/

plantuml

The package [`font-noto-cjk`](https://pkgs.alpinelinux.org/package/edge/community/x86_64/font-noto-cjk) (and all its dependencies) is quite large, approximately 200MB.
Here's the Docker image size before (0.10.0) and after (latest):

```
yuzutech/kroki     0.10.0     434MB
yuzutech/kroki     latest     650MB
```

resolves #386

### DIFF
--- a/server/ops/docker/jdk11-alpine/Dockerfile
+++ b/server/ops/docker/jdk11-alpine/Dockerfile
@@ -21,10 +21,7 @@ RUN apk add --update --no-cache \
            giflib-dev \
            graphviz \
            ttf-freefont \
-# Chinese font (available only in edge/testing)
- && apk add --update --no-cache \
-            --repository https://dl-cdn.alpinelinux.org/alpine/edge/testing/ \
-           wqy-zenhei
+           font-noto-cjk
 
 COPY --chown=kroki:kroki ops/docker/logback.xml /etc/kroki/logback.xml
 


zengxs:
The package `font-noto-cjk-extra` includes all font weights, but most of them are unnecessary.

I think `source-han-sans` might be a better choice:
(`noto-cjk` is same with `source-han-sans`, ref [this](https://github.com/adobe-fonts/source-han-sans/issues/122#issuecomment-128087438))
| Weight | Url | Size |
|:-------|:----|:----|
| Regular | <https://github.com/adobe-fonts/source-han-sans/blob/release/OTC/SourceHanSans-Regular.ttc> | 19.1 MB |
| Normal | <https://github.com/adobe-fonts/source-han-sans/blob/release/OTC/SourceHanSans-Normal.ttc> | 17.8 MB |
| Medium | <https://github.com/adobe-fonts/source-han-sans/blob/release/OTC/SourceHanSans-Medium.ttc> | 18 MB |

I think the weight `Regular` is enough for kroki.

And if this font is installed, the font `wqy-zenhei` (Chinese font) can be removed.

Mogztter:
Thanks for your input @zengxs
That's exactly the kind of information I was looking for! 👍🏻 

`source-han-sans` is not available as an apk package but I will try to manually install the font using TrueType Collection (ttc) files.

Mogztter:
I took a closer look and Noto has exactly the same size:

| Weight | Url | Size |
|:-------|:----|:----|
| Regular | <https://github.com/googlefonts/noto-cjk/blob/main/NotoSansCJK-Regular.ttc> | 19.1 MB |
| Medium | <https://github.com/googlefonts/noto-cjk/blob/main/NotoSansCJK-Medium.ttc> | 18 MB |

The package `font-noto-cjk-extra` is quite large because it contains several variations:

- NotoSansCJK-Bold.ttc
- NotoSansCJK-Regular.ttc
- NotoSerifCJK-Bold.ttc
- NotoSerifCJK-Regular.ttc
- NotoSansCJK-Black.ttc
- NotoSansCJK-DemiLight.ttc
- NotoSansCJK-Light.ttc
- NotoSansCJK-Medium.ttc
- NotoSansCJK-Thin.ttc
- NotoSerifCJK-Black.ttc
- NotoSerifCJK-ExtraLight.ttc
- NotoSerifCJK-Light.ttc
- NotoSerifCJK-Medium.ttc
- NotoSerifCJK-SemiBold.ttc

As you've mentioned, we probably don't want to include all font weights.
I just noticed that we are using the `-extra` package but instead we could use: `font-noto-cjk` which only contains:

- NotoSansCJK-Bold.ttc
- NotoSansCJK-Regular.ttc
- NotoSerifCJK-Bold.ttc
- NotoSerifCJK-Regular.ttc

> I think source-han-sans might be a better choice:

I don't have a strong opinion since I can't tell if `source-han-sans` is *objectively* better than `noto-sans` or not.

Mogztter:
Here's the Docker image size before (0.11.0) and after (latest) with `font-noto-cjk` package:

```
yuzutech/kroki     0.11.0     455MB
yuzutech/kroki     latest     530MB
```

That's a 75MB increase which is fine and I'm willing to move forward with this change.

zengxs:
`noto-sans-cjk` is also good, it is default cjk font in most linux desktop distros, but `source-han-sans` is more famous in China.

In fact, they are almost the same font (ref [this post](https://developers.googleblog.com/2014/07/noto-cjk-font-that-is-complete.html)), just with different names.

Mogztter:
Thanks again for your feedback and links to resources, this is greatly appreciated 👍🏻 

> noto-sans-cjk is also good, it is default cjk font in most linux desktop distros, but source-han-sans is more famous in China.
In fact, they are almost the same font (ref this post), just with different names.

OK, I see.
It seems that Google and Adobe worked hand in hand to produce a CJK font.

> If you want to install font from alpine package repository, noto-sans is the best choice.
If you want to install font manually, source-han-sans's font-name has better compatibility than noto-sans (many programs/websites optimized for source-han-sans but noto-sans not), although this improvement is negligible.

I quite like the fact that `noto-sans` is available as an apk package, it keeps the Dockerfile simple.
Let's start with Noto and see if this option is good enough or not (i.e., we can always reevaluate later).

Mogztter:
For reference, here's the result using Noto:

noto

PR #613, opened 2021-02-24
Dockerfile with user kroki and kroki as owner

lind:
User and group added and home dir as workdir.

### DIFF
--- a/blockdiag/Dockerfile
+++ b/blockdiag/Dockerfile
@@ -1,21 +1,25 @@
 FROM python:3.8.7-alpine3.11
 
-WORKDIR /usr/src/app
+RUN addgroup -g 1000 kroki && adduser -D -G kroki -u 1000 kroki
+
+# add runtime dependency
+RUN apk --update --no-cache add jpeg zlib fontconfig ttf-dejavu
 
 ENV LIBRARY_PATH=/lib:/usr/lib
 
-COPY requirements.txt /requirements.txt
+WORKDIR /usr/local/kroki
 
-# add runtime dependency
-RUN apk --update add jpeg zlib fontconfig ttf-dejavu
+COPY --chown=kroki:kroki requirements.txt .
 
 # build
-RUN apk --update add --virtual build-dependencies build-base python3-dev jpeg-dev zlib-dev\
-  && pip install --no-cache-dir -r /requirements.txt \
+RUN apk --update --no-cache add --virtual build-dependencies build-base python3-dev jpeg-dev zlib-dev \
+  && pip install --no-cache-dir -r requirements.txt \
   && apk del build-dependencies
 
-COPY src .
+COPY --chown=kroki:kroki src .
 
 EXPOSE 8001
+
+USER kroki
 
 CMD ["gunicorn", "--preload", "--bind", "0.0.0.0:8001", "wsgi"]
--- a/bpmn/Dockerfile
+++ b/bpmn/Dockerfile
@@ -26,13 +26,18 @@ RUN /usr/local/bin/pkg --targets ${NODE}-${PLATFORM}-${ARCH} . -o app.bin
 # Create the image
 FROM alpine:3.13.2
 
-COPY --from=builder /app/app.bin /node/bin/bpmn
-COPY assets /app/assets
-ENV KROKI_BPMN_PAGE_URL=file:///app/assets/index.html
+RUN addgroup -g 1000 kroki && adduser -D -G kroki -u 1000 kroki
 
 RUN apk add --no-cache --update chromium
+
+COPY --from=builder /app/app.bin /usr/bin/bpmn
+COPY --chown=kroki:kroki assets /usr/local/kroki/assets
+
+ENV KROKI_BPMN_PAGE_URL=file:///usr/local/kroki/assets/index.html
 ENV PUPPETEER_EXECUTABLE_PATH=/usr/lib/chromium/chrome
 
 EXPOSE 8003
 
-ENTRYPOINT ["/node/bin/bpmn"]
+USER kroki
+
+ENTRYPOINT ["/usr/bin/bpmn"]
--- a/excalidraw/Dockerfile
+++ b/excalidraw/Dockerfile
@@ -26,13 +26,18 @@ RUN /usr/local/bin/pkg --targets ${NODE}-${PLATFORM}-${ARCH} . -o app.bin
 # Create the image
 FROM alpine:3.12.0
 
-COPY --from=builder /app/app.bin /node/bin/excalidraw
-COPY assets /app/assets
-ENV KROKI_EXCALIDRAW_PAGE_URL=file:///app/assets/index.html
+RUN addgroup -g 1000 kroki && adduser -D -G kroki -u 1000 kroki
 
 RUN apk add --no-cache --update chromium
+
+COPY --from=builder /app/app.bin /usr/bin/excalidraw
+COPY --chown=kroki:kroki assets /usr/local/kroki/assets
+
+ENV KROKI_EXCALIDRAW_PAGE_URL=file:///usr/local/kroki/assets/index.html
 ENV PUPPETEER_EXECUTABLE_PATH=/usr/lib/chromium/chrome
 
 EXPOSE 8004
 
-ENTRYPOINT ["/node/bin/excalidraw"]
+USER kroki
+
+ENTRYPOINT ["/usr/bin/excalidraw"]
--- a/mermaid/Dockerfile
+++ b/mermaid/Dockerfile
@@ -26,11 +26,9 @@ RUN /usr/local/bin/pkg --targets ${NODE}-${PLATFORM}-${ARCH} . -o app.bin
 # Create the image
 FROM alpine:3.13.2
 
-RUN apk add --no-cache --update chromium
-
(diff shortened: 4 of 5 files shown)


Mogztter:
Thanks, as soon as i get the chance I will review your contribution :+1:

bpmn/Dockerfile
```diff
@@ -26,13 +26,20 @@ RUN /usr/local/bin/pkg --targets ${NODE}-${PLATFORM}-${ARCH} . -o app.bin
 # Create the image
 FROM alpine:3.13.2
 
-COPY --from=builder /app/app.bin /node/bin/bpmn
-COPY assets /app/assets
+RUN apk add --no-cache --update chromium
+
+ENV USER_ID=1000 USER=kroki
```
Mogztter, line 31: I don't think we should use environment variables, I'm fine with small "code duplication".
As far as I know, the environment variables declared with `ENV` will persist when a container is run from the resulting image.
In this case, `USER_ID` and `USER` are build-only constants.

lind: 👍  Agree. I'll change.

lind:
@Mogztter I'm not sure about the failing [smokeTests](https://github.com/yuzutech/kroki/pull/613/checks?check_run_id=1982252830) and if it comes from the Dockerfile changes?

bpmn/Dockerfile
```diff
@@ -26,13 +26,19 @@ RUN /usr/local/bin/pkg --targets ${NODE}-${PLATFORM}-${ARCH} . -o app.bin
...
-COPY assets /app/assets
+RUN apk add --no-cache --update chromium
+
+RUN addgroup -g 1000 kroki && adduser -D -G kroki -u 1000 kroki
+
+USER kroki
+WORKDIR /home/kroki
+
+COPY --from=builder /app/app.bin /usr/bin/bpmn
+COPY --chown=kroki:kroki assets /app/assets
```
lind, line 37: Where should `assets` be placed? Keep `/app/assets`?

Mogztter:
> I'm not sure about the failing smokeTests and if it comes from the Dockerfile changes?

I don't think it's related since I can reproduce this issue on master. Maybe an intermittent failure on GitHub Actions... but it's really weird because the error is exactly the same on every pull request.
The server responds to 5 requests and then the connection is reset (`ECONNRESET`) and then the server is presumably dead/unavailable (`ECONNREFUSED`)... :thinking:

bpmn/Dockerfile
Mogztter: I think `/usr/local/kroki/assets` would be fine, so we know where to look for Kroki related files on all Docker images.
Don't forget to update `KROKI_BPMN_PAGE_URL`.

bpmn/Dockerfile
```diff
@@ -26,13 +26,19 @@ RUN /usr/local/bin/pkg --targets ${NODE}-${PLATFORM}-${ARCH} . -o app.bin
...
 FROM alpine:3.13.2
 
-COPY --from=builder /app/app.bin /node/bin/bpmn
-COPY assets /app/assets
+RUN apk add --no-cache --update chromium
+
+RUN addgroup -g 1000 kroki && adduser -D -G kroki -u 1000 kroki
+
+USER kroki
+WORKDIR /home/kroki
```
Mogztter, line 34: I wonder if `WORKDIR` is needed here? We are only using absolute path so nothing gets added in `/home/kroki`?

lind:
Still failing on smokeTests with `ECONNREFUSED`. I'm afraid I'm not that much help finding the cause of that...

Works fine locally running `docker-compose up -d`...

Mogztter:
> Works fine locally running docker-compose up -d...

Yes... I cannot reproduce this issue either.
Running `make smokeTests` on my machine is working fine.
Even if it's unrelated, I would like to fix this issue before merging your great contribution otherwise it will be even harder to find the root cause.

I will try to retrieve logs from all running containers maybe it will give us an hint.

blockdiag/Dockerfile
```diff
@@ -16,6 +16,9 @@ RUN apk --update add --virtual build-dependencies build-base python3-dev jpeg-de
 
 COPY src .
 
+RUN addgroup -g 1000 kroki && adduser -D -G kroki -u 1000 kroki
```
Mogztter, line 19: Could you please move this statement just after `FROM`?
The reason is that some packages could potentially add a new user/group and take id 1000.

Mogztter:
Could you please rebase on master (to include this fix https://github.com/yuzutech/kroki/commit/7d15fa17e7571bf00d82bab9b29923fd2b8c87ae) to get rid of the annoying failure?

lind:
> Could you please rebase on master (to include this fix [7d15fa1](https://github.com/yuzutech/kroki/commit/7d15fa17e7571bf00d82bab9b29923fd2b8c87ae)) to get rid of the annoying failure?

I hope I did the rebase correct... The smoke tests is now green at least

Mogztter:
> I hope I did the rebase correct... The smoke tests is now green at least

Looks fine. I rebased and squashed commits.
Could you please review it to make sure that I didn't mess up?

excalidraw/Dockerfile
```diff
@@ -26,13 +26,18 @@ RUN /usr/local/bin/pkg --targets ${NODE}-${PLATFORM}-${ARCH} . -o app.bin
...
+COPY --from=builder /app/app.bin /usr/bin/excalidraw
+COPY --chown=kroki:kroki assets /usr/local/kroki/assets
+
+ENV KROKI_EXCALIDRAW_PAGE_URL=file:///usr/local/kroki/assets/index.html
 ENV PUPPETEER_EXECUTABLE_PATH=/usr/lib/chromium/chrome
 
 EXPOSE 8004
 
-ENTRYPOINT ["/node/bin/excalidraw"]
+USER kroki
```
lind, line 41: @Mogztter I think I have to move `USER` to before `ENV` otherwise the `ENV` is not registered on the kroki user. Or am I wrong? 

The same for the main Docker file...

lind: No it seems the `ENV` is there for the kroki user as well...

lind:
> > I hope I did the rebase correct... The smoke tests is now green at least
> 
> Looks fine. I rebased and squashed commits.
> Could you please review it to make sure that I didn't mess up?

It seems fine to me!

mermaid/Dockerfile
```diff
@@ -26,13 +26,18 @@ RUN /usr/local/bin/pkg --targets ${NODE}-${PLATFORM}-${ARCH} . -o app.bin
 # Create the image
 FROM alpine:3.13.2
 
-COPY --from=builder /app/app.bin /node/bin/mermaid
-COPY assets /app/assets
-ENV KROKI_MERMAID_PAGE_URL=file:///app/assets/index.html
-
 RUN apk add --no-cache --update chromium
+
+RUN addgroup -g 1000 kroki && adduser -D -G kroki -u 1000 kroki
```
Mogztter, line 31:
```suggestion
RUN addgroup -g 1000 kroki && adduser -D -G kroki -u 1000 kroki

RUN apk add --no-cache --update chromium
```

mermaid/Dockerfile
```diff
@@ -26,13 +26,18 @@ RUN /usr/local/bin/pkg --targets ${NODE}-${PLATFORM}-${ARCH} . -o app.bin
...
+
+USER kroki
```
Mogztter, line 33:
```suggestion
```

mermaid/Dockerfile
```diff
@@ -26,13 +26,18 @@ RUN /usr/local/bin/pkg --targets ${NODE}-${PLATFORM}-${ARCH} . -o app.bin
...
+COPY --from=builder /app/app.bin /usr/bin/mermaid
+COPY --chown=kroki:kroki assets /usr/local/kroki/assets
+ENV KROKI_MERMAID_PAGE_URL=file:///usr/local/kroki/assets/index.html
+
 ENV PUPPETEER_EXECUTABLE_PATH=/usr/lib/chromium/chrome
 
 EXPOSE 8002
 
-ENTRYPOINT ["/node/bin/mermaid"]
+ENTRYPOINT ["/usr/bin/mermaid"]
```
Mogztter, line 43:
```suggestion
USER kroki

ENTRYPOINT ["/usr/bin/mermaid"]
```

mermaid/Dockerfile
```diff
@@ -26,13 +26,18 @@ RUN /usr/local/bin/pkg --targets ${NODE}-${PLATFORM}-${ARCH} . -o app.bin
...
-COPY assets /app/assets
-ENV KROKI_MERMAID_PAGE_URL=file:///app/assets/index.html
+RUN addgroup -g 1000 kroki && adduser -D -G kroki -u 1000 kroki
 
 RUN apk add --no-cache --update chromium
+
+COPY --from=builder /app/app.bin /usr/bin/mermaid
+COPY --chown=kroki:kroki assets /usr/local/kroki/assets
+ENV KROKI_MERMAID_PAGE_URL=file:///usr/local/kroki/assets/index.html
+
```
Mogztter, line 36:
```suggestion

ENV KROKI_MERMAID_PAGE_URL=file:///usr/local/kroki/assets/index.html
```

Mogztter:
And merged! Thanks @lind

PR #616, opened 2021-02-27
Document how to to deploy on Kubernetes

Mogztter:


Mogztter:
Many thanks @teochenglim :+1:

teochenglim:
No problem. I learned about this framework because our developers are using it. And internally we deployed this to our EKS cluster. It is about the same yaml anyway so I thought it would be nice for you to have it too.

Mogztter:
Definitely, I believe it will help a lot of users :ok_hand: 
The documentation is now live: https://docs.kroki.io/kroki/setup/install/#_using_kubernetes :tada:

Issue #555, opened 2021-01-19
Make the max URI length configurable on Vert.x (using configuration)


mariotoffia:
Hi and thanks for this great piece of software! :)

I'm using excalidraw via asciidoc-kroki (from @Mogztter) - It seems that if the file gets to big or too many objects in the drawing it will fail to render the svg image. Attached both _JSON_ files respective images of failing and the one working.

This works:
![image](https://user-images.githubusercontent.com/1716527/105049300-cd2a4780-5a6c-11eb-9644-8839d7f62224.png)

This (from excalidraw.com) do not work:
![image](https://user-images.githubusercontent.com/1716527/105049414-e7fcbc00-5a6c-11eb-965f-52d4a744773b.png)

[json-files.zip](https://github.com/yuzutech/kroki/files/5836530/json-files.zip)

Cheers,
 Mario :)

Mogztter:
> I'm using excalidraw via asciidoc-kroki (from @Mogztter) - It seems that if the file gets to big or too many objects in the drawing it will fail to render the svg image. Attached both JSON files respective images of failing and the one working.

The public instance of [Kroki](https://kroki.io) returns a "414 Request-URI Too Large" when the URI exceeds 4K bytes.
It ensures that the public instance will run smoothly.
Please note that you can configure a higher value if you are self-hosting Kroki.

In this case, the Asciidoctor Kroki extension should use a POST request but I think the limit should be 4000 and not 4096.

https://github.com/Mogztter/asciidoctor-kroki/blob/35f75e232a5671fe4e404a9b6e4516e011b5919e/src/kroki-client.js#L26

I will transfer this issue since it's not related to Kroki.

Mogztter:
> I will transfer this issue since it's not related to Kroki (or at least Kroki is working as expected).

Actually I cannot transfer an issue from another organization :disappointed: 
Could you please open a new issue at: https://github.com/mogztter/asciidoctor-kroki?

mariotoffia:
Thanks @Mogztter - I'll create a new issue at asciidoctor-kroki!

Mogztter:
Currently, this value is not configurable. For reference, we could update the value using https://vertx.io/docs/apidocs/io/vertx/core/http/HttpServerOptions.html#setMaxInitialLineLength-int-.

It would allow to increase the default value (4096) on self-hosted instance.

lind:
A proposal for this issue in https://github.com/yuzutech/kroki/pull/626

[...]

PR #626, opened 2021-03-06
Make the max URI length configurable

lind:
Proposal for https://github.com/yuzutech/kroki/issues/555

### DIFF
--- a/docs/modules/setup/pages/configuration.adoc
+++ b/docs/modules/setup/pages/configuration.adoc
@@ -185,4 +185,4 @@ NOTE: If you are using the default `docker-compose.yaml` file you can rely on th
 Some diagrams, like Excalidraw, have verbose textual descriptions that will produce long URI.
 If the URI requested by the client is longer than the server is willing to interpret, the server will return a 414 (Request-URI Too Long) response status code.
 The https://netty.io/4.0/api/io/netty/handler/codec/http/HttpRequestDecoder.html#HttpRequestDecoder--[default max URI length] in Vert.x is 4096.
-This could be changed by setting `KROKI_MAX_LINE_LENGTH` environment variable.
+You can update this default value by setting `KROKI_MAX_URI_LENGTH` environment variable.
--- a/docs/modules/setup/pages/configuration.adoc
+++ b/docs/modules/setup/pages/configuration.adoc
@@ -184,5 +184,5 @@ NOTE: If you are using the default `docker-compose.yaml` file you can rely on th
 
 Some diagrams, like Excalidraw, have verbose textual descriptions that will produce long URI.
 If the URI requested by the client is longer than the server is willing to interpret, the server will return a 414 (Request-URI Too Long) response status code.
-The https://netty.io/4.0/api/io/netty/handler/codec/http/HttpRequestDecoder.html#HttpRequestDecoder--[default length] of the max URI length is 4096.
+The https://netty.io/4.0/api/io/netty/handler/codec/http/HttpRequestDecoder.html#HttpRequestDecoder--[default max URI length] in Vert.x is 4096.
 This could be changed by setting `KROKI_MAX_LINE_LENGTH` environment variable.
--- a/docs/modules/setup/pages/configuration.adoc
+++ b/docs/modules/setup/pages/configuration.adoc
@@ -182,6 +182,7 @@ NOTE: If you are using the default `docker-compose.yaml` file you can rely on th
 
 == Max URI length
 
-Some diagrams, like Excalidraw, could have longer URI length then the server accepts and therefore returns with a 414 (Request-URI Too Long) in the response.
+Some diagrams, like Excalidraw, have verbose textual descriptions that will produce long URI.
+If the URI requested by the client is longer than the server is willing to interpret, the server will return a 414 (Request-URI Too Long) response status code.
 The https://netty.io/4.0/api/io/netty/handler/codec/http/HttpRequestDecoder.html#HttpRequestDecoder--[default length] of the max URI length is 4096.
 This could be changed by setting `KROKI_MAX_LINE_LENGTH` environment variable.
--- a/docs/modules/setup/pages/configuration.adoc
+++ b/docs/modules/setup/pages/configuration.adoc
@@ -187,5 +187,6 @@ If the URI requested by the client is longer than the server is willing to inter
 The https://netty.io/4.0/api/io/netty/handler/codec/http/HttpRequestDecoder.html#HttpRequestDecoder--[default max URI length] in Vert.x is 4096.
 You can update this default value by setting `KROKI_MAX_URI_LENGTH` environment variable.
 
-TIP: The most browsers support >64k, also newer Edge, but that's a bit excessive so start with lower value.
-If you're supporting Edge 16 then the max value should be 10240, for IE 11 not more than 5120.
+TIP: Keep in mind that browsers also have a URI limit on `<img>` tags.
+Most modern browsers [support a URI length greater than 64000](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers/417184#417184) on `<img>` tags but this value is probably a bit excessive.
+We recommend to use a maximum length that's not greater than 8192 and not greater than 5120 if you are supporting IE 11.
--- a/docs/modules/setup/pages/configuration.adoc
+++ b/docs/modules/setup/pages/configuration.adoc
@@ -179,3 +179,9 @@ KROKI_EXCALIDRAW_HOST:: Host of the Excalidraw container (default: `127.0.0.1`).
 KROKI_EXCALIDRAW_PORT:: Port of the Excalidraw container (default: `8004`).
 
 NOTE: If you are using the default `docker-compose.yaml` file you can rely on the default values.
+
+== Max URI length
+
+Some diagrams, like Excalidraw, could have longer URI length then the server accepts and therefore returns with a 414 (Request-URI Too Long) in the response.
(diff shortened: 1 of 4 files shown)


docs/modules/setup/pages/configuration.adoc
```diff
@@ -179,3 +179,9 @@ KROKI_EXCALIDRAW_HOST:: Host of the Excalidraw container (default: `127.0.0.1`).
 KROKI_EXCALIDRAW_PORT:: Port of the Excalidraw container (default: `8004`).
 
 NOTE: If you are using the default `docker-compose.yaml` file you can rely on the default values.
+
+== Max URI length
+
+Some diagrams, like Excalidraw, could have longer URI length then the server accepts and therefore returns with a 414 (Request-URI Too Long) in the response.
```
Mogztter, line 185:
```suggestion
Some diagrams, like Excalidraw, have verbose textual descriptions that will produce long URI.
If the URI requested by the client is longer than the server is willing to interpret, the server will return a 414 (Request-URI Too Long) response status code.
```

docs/modules/setup/pages/configuration.adoc
```diff
@@ -179,3 +179,9 @@ KROKI_EXCALIDRAW_HOST:: Host of the Excalidraw container (default: `127.0.0.1`).
...
+The https://netty.io/4.0/api/io/netty/handler/codec/http/HttpRequestDecoder.html#HttpRequestDecoder--[default length] of the max URI length is 4096.
```
Mogztter, line 186: The https://netty.io/4.0/api/io/netty/handler/codec/http/HttpRequestDecoder.html#HttpRequestDecoder--[default max URI length] in Vert.x is 4096.

docs/modules/setup/pages/configuration.adoc
```diff
@@ -179,3 +179,9 @@ KROKI_EXCALIDRAW_HOST:: Host of the Excalidraw container (default: `127.0.0.1`).
...
+This could be changed by setting `KROKI_MAX_LINE_LENGTH` environment variable.
```
Mogztter, line 187:
```suggestion
You can update this default value by setting `KROKI_MAX_LINE_LENGTH` environment variable.
```

server/src/main/java/io/kroki/server/Server.java
```diff
@@ -68,7 +69,14 @@ public void start(Promise<Void> startPromise) {
   }
 
   static void start(Vertx vertx, JsonObject config, Handler<AsyncResult<HttpServer>> listenHandler) {
-    HttpServer server = vertx.createHttpServer();
+    Integer maxInitialLineLength = config.getInteger("KROKI_MAX_LINE_LENGTH");
```
Mogztter, line 72: I think we should use the name `KROKI_MAX_URI_LENGTH`.
In my opinion, max line is a bit unclear.

server/src/main/java/io/kroki/server/Server.java
```diff
@@ -68,7 +69,14 @@ public void start(Promise<Void> startPromise) {
...
+    HttpServerOptions serverOptions = new HttpServerOptions();
+    if (maxInitialLineLength != null) {
+      logger.info("KROKI_MAX_LINE_LENGTH: {}", maxInitialLineLength);
```
Mogztter, line 75: It could be useful to add a _debug_ log with the whole configuration but I don't think we should log individual option using info level.
Could you please remove this line?

```suggestion
```

Mogztter:
Thanks @lind!
Overall it's looking good, I'm very pleased that you've added tests :+1: 

I left a few comments, let me know what you think.

server/src/main/java/io/kroki/server/Server.java
lind: Agree

server/src/main/java/io/kroki/server/Server.java
lind: Sure. I think logging of config is good to be able to collect config from many containers and reported by the log indexer but as you mention that should in that case ber done for the whole configuration.

And maybe not needed for this app.

docs/modules/setup/pages/configuration.adoc
```diff
@@ -179,3 +179,10 @@ KROKI_EXCALIDRAW_HOST:: Host of the Excalidraw container (default: `127.0.0.1`).
 KROKI_EXCALIDRAW_PORT:: Port of the Excalidraw container (default: `8004`).
 
 NOTE: If you are using the default `docker-compose.yaml` file you can rely on the default values.
+
+== Max URI length
+
+Some diagrams, like Excalidraw, have verbose textual descriptions that will produce long URI.
```
lind, line 185: @Mogztter I was thinking about adding some information about what length that could be added byt figured out that "best practice" changes dependent on what browsers your aiming for ref [maximum-length-of-a-url-in-different-browsers](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers/417184#417184). But do you think something about this should be mentioned?

Mogztter: Definitely, that's a good idea :+1: 
I remember looking for this information so a `TIP: ` would be extremely valuable.
In a nutshell, if you don't need to navigate to the URL (i.e., you only want to display the image using the `` tag) and you don't need to support IE than you can opt for a max URI length of 10240.
And If you ignore Edge, then you can use >64k but I wouldn't recommend this value (it seems excessive).

lind: 👍 I'll make a proposal

docs/modules/setup/pages/configuration.adoc
lind: Changed

docs/modules/setup/pages/configuration.adoc
lind: Have a look on the new commit with the tip

docs/modules/setup/pages/configuration.adoc
```diff
@@ -179,3 +179,13 @@ KROKI_EXCALIDRAW_HOST:: Host of the Excalidraw container (default: `127.0.0.1`).
...
+If the URI requested by the client is longer than the server is willing to interpret, the server will return a 414 (Request-URI Too Long) response status code.
+The https://netty.io/4.0/api/io/netty/handler/codec/http/HttpRequestDecoder.html#HttpRequestDecoder--[default max URI length] in Vert.x is 4096.
+You can update this default value by setting `KROKI_MAX_URI_LENGTH` environment variable.
+
+TIP: The most browsers support >64k, also newer Edge, but that's a bit excessive so start with lower value.
+If you're supporting Edge 16 then the max value should be 10240, for IE 11 not more than 5120.
```
Mogztter, line 191: What do you think about:

```suggestion
TIP: Keep in mind that browsers also have a URI limit on `<img>` tags.
Most modern browsers [support a URI length greater than 64000](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers/417184#417184) on `<img>` tags but this value is probably a bit excessive.
We recommend to use a maximum length that's not greater than 8192 and not greater than 5120 if you are supporting IE 11.
```

server/src/main/java/io/kroki/server/Server.java
```diff
@@ -68,7 +69,13 @@ public void start(Promise<Void> startPromise) {
...
 
   static void start(Vertx vertx, JsonObject config, Handler<AsyncResult<HttpServer>> listenHandler) {
-    HttpServer server = vertx.createHttpServer();
+    Integer maxInitialLineLength = config.getInteger("KROKI_MAX_URI_LENGTH");
+    HttpServerOptions serverOptions = new HttpServerOptions();
+    if (maxInitialLineLength != null) {
+      serverOptions.setMaxInitialLineLength(maxInitialLineLength);
+    }
+
+    HttpServer server = vertx.createHttpServer(serverOptions);
```
Mogztter, line 78: I've renamed the variable name to `maxUriLength`:

```suggestion
    Integer maxUriLength = config.getInteger("KROKI_MAX_URI_LENGTH");
    HttpServerOptions serverOptions = new HttpServerOptions();
    if (maxUriLength != null) {
      serverOptions.setMaxInitialLineLength(maxUriLength);
    }
    HttpServer server = vertx.createHttpServer(serverOptions);
```

Mogztter:
Final round and then we are good to go! :tada:

docs/modules/setup/pages/configuration.adoc
```diff
@@ -179,3 +179,14 @@ KROKI_EXCALIDRAW_HOST:: Host of the Excalidraw container (default: `127.0.0.1`).
...
+
+== Max URI length
+
+Some diagrams, like Excalidraw, have verbose textual descriptions that will produce long URI.
+If the URI requested by the client is longer than the server is willing to interpret, the server will return a 414 (Request-URI Too Long) response status code.
+The https://netty.io/4.0/api/io/netty/handler/codec/http/HttpRequestDecoder.html#HttpRequestDecoder--[default max URI length] in Vert.x is 4096.
+You can update this default value by setting `KROKI_MAX_URI_LENGTH` environment variable.
+
+TIP: Keep in mind that browsers also have a URI limit on `<img>` tags.
+Most modern browsers [support a URI length greater than 64000](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers/417184#417184) on `<img>` tags but this value is probably a bit excessive.
```
Mogztter, line 191:
```suggestion
Most modern browsers https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers/417184#417184[support a URI length greater than 64000] on `<img>` tags but this value is probably a bit excessive.
```

Mogztter:
@lind And merged! Thanks again :smile:

PR #668, opened 2021-03-29
adding support for bit-field

am21:
Here are my changes to add support for bitfield (bit-field is the name for npm).
I know very little about JavaScript, you will want to review my changes :-)

I have fixed asciidoctor-kroki (PR coming up soon) and it works really well for me.

Merci!


wavedrom/package.json
```diff
@@ -17,7 +17,8 @@
   "dependencies": {
     "json5": "2.2.0",
     "onml": "2.1.0",
-    "wavedrom": "2.6.8"
+    "wavedrom": "2.6.8",
+    "bit-field": "1.3.6"
```
Mogztter, line 21: `bit-field` is already a dependency of `wavedrom`, see: https://github.com/wavedrom/wavedrom/blob/7b78999021378bf49ebc1dafa31baf3d09f20a64/package.json#L61

am21: So, I guess there is nothing to do then? I'm confused now. Maybe it didn't work because the yuzutech/kroki container I pulled has an older version of wavedrom?

Mogztter: Probably, the current Docker image (0.11) is using WaveDrom 2.3.2.

Do you have a sample diagram so we can add a test to make sure that it's working on the master branch with WaveDrom 2.6.8 ?

am21: Yes, here is one, taken from https://observablehq.com/@drom/bitfield-testbox?collection=@drom/bitfield:

```
{reg: [
    {name: "IPO",   bits: 8, attr: "RO"},
    {               bits: 7},
    {name: "BRK",   bits: 5, attr: ["RW"], type: 4},
    {name: "CPK",   bits: 1},
    {name: "Clear", bits: 3, type: 5},
    {               bits: 8}
]}
```

Mogztter: This example is already working today with Kroki 0.11:

image

am21: This simple one with rotate should do it:
```
{reg: [
  {name: "long_label", bits: 1, rotate: -90},
  {name: '"longer_label", bits: 3, rotate: -45},
  {bits: 28}
],
config: {vspace: 150}
}
```

Mogztter: Thanks, I can confirm that this definition does not work on Kroki 0.11, I will try the same definition against master branch (WaveDrom 2.6.8)

Mogztter:
Confirmed in #676, bitfield (and `rotate`) will be supported in the next release of Kroki.

I think we can close this pull request.

am21:
Yes, definitely.

PR #700, opened 2021-04-18
tell GCC to build a static binary for pikchr

jerojasro:
the previous version of this dockerfile didn't build a static binary, but that issue was masked because the image used to build the binary, and the image where the binary ran, both have the required dynamic libraries.

This commit fixes the issue, by passing GCC the `-static` flag to build a static binary.


Mogztter:
Nice catch, thanks @jerojasro 👍🏻

Issue #718, opened 2021-05-05
PlantUML C4 support from behind firewall


lind:
I get an error using C4 in PlantUML from behind a firewall. It's a 400 `Cannot open URL (line: 504)` but in the log it's also a:
`issue raw.githubusercontent.com java.util.concurrent.ExecutionException: javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target` 
This log is probably from PlantUML itself...?

If I downgrade Kroki to 0.11.0 I don't get the exception so I wonder if it is related to the [include](https://github.com/yuzutech/kroki/blob/main/server/src/main/resources/c4_component.puml#L5) in the C4 files introduced in https://github.com/yuzutech/kroki/issues/675

Is the `include` needed? It's not present in [the 0.11 version](https://github.com/yuzutech/kroki/tree/v0.11.0/server/src/main/resources)

I tried to set env.var: `RELATIVE_INCLUDE` to . but with the same error.

It might be related to this issue: https://github.com/yuzutech/kroki/issues/578

Mogztter:
@lind Could you please share a simple diagram definition that reproduces this error? Depending on how you include C4, Kroki will either to delegate to PlantUML or use local files.

> Cannot open URL (line: 504) but in the log it's also a: issue raw.githubusercontent.com java.util.concurrent.ExecutionException: javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target

In any case, I wasn't expecting PlantUML to do a remote call...

lind:
[...]

PR #725, opened 2021-05-07
remove the include part in c4 .puml files

lind:
To resolve https://github.com/yuzutech/kroki/issues/718

### DIFF
--- a/server/src/main/resources/c4_component.puml
+++ b/server/src/main/resources/c4_component.puml
@@ -1,10 +1,3 @@
-' convert it with additional command line argument -DRELATIVE_INCLUDE="." to use locally
-!if %variable_exists("RELATIVE_INCLUDE")
-  !include %get_variable_value("RELATIVE_INCLUDE")/C4_Container.puml
-!else
-  !include https://raw.githubusercontent.com/plantuml-stdlib/C4-PlantUML/master/C4_Container.puml
-!endif
-
 ' Scope: A single container.
 ' Primary elements: Components within the container in scope.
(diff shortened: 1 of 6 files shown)


lind:
@Mogztter Some whitespaces removed as well

Mogztter:
Looks good!
It would be great to add a test case to make sure that we don't rely on the network but it might be a bit tricky to test... 🤔 
I will give it some thought... maybe we could set `socksProxyHost` and `socksProxyPort` to a non existing proxy using `System.setProperty`?

Mogztter:
> It would be great to add a test case to make sure that we don't rely on the network but it might be a bit tricky to test... thinking
I will give it some thought... maybe we could set socksProxyHost and socksProxyPort to a non existing proxy using System.setProperty?

Seems to be working, I've added a test

Mogztter:
Thanks for the fix 👏🏻

lind:
Thanks for fast response 😄 

@Mogztter Do you think this might come with a 12.2 release some time soon? It would be nice to have C4 working again with [larger Excalidraw diagrams](https://docs.kroki.io/kroki/setup/configuration/#_max_uri_length)...

Mogztter:
The main branch already contains new "features", so it will probably be version 0.13.0 but yes I'm trying to do monthly releases.

> It would be nice to have C4 working again and larger Excalidraw diagrams.

This is already available in Kroki 0.12.0, if you are using your own instance you can use `KROKI_MAX_URI_LENGTH` to increase the max URI length.

I might increase it on the public instance.... I'm still not decided but I think I'm gonna increase it to 8192 and see what happens... if it's too much then I might revert it to the default value 4096.

PR #737, opened 2021-05-09
resolves #698 update the description on Docker Hub

### LINKED ISSUE #698: Automatically update the description on hub.docker.com
The description on hub.docker.com is a bit outdated. We could probably use https://github.com/peter-evans/dockerhub-description to automatically update it from a README.md.

lind:
A proposal for https://github.com/yuzutech/kroki/issues/698

I'm new to GitHub Actions but the [action](https://github.com/peter-evans/dockerhub-description) mentioned in the https://github.com/yuzutech/kroki/issues/698 should work. But I don't know if there is a way to test this without actually change the description in Dockerhub.

I''m not sure the README is the right file to use as the Dockerhub description since the README includes build information etc. not relevant to Dockerhub users and also there are a need for multiple Dockerhub descriptions, one for each docker image; bmpn, excalidraw...

Proposes one description file for each image in the same dir as the Dockerfile `/bpmn`, `/excalidraw`...

The proposed DOCKERHUB.md is just a copy and slight simplification of the description on Dockerhub today and needs some more work.

What do you think @Mogztter


.github/workflows/dockerhub-description.yml
```diff
@@ -0,0 +1,21 @@
+on:
+  push:
+    branches:
+      - master
```
Mogztter, line 4: I've renamed the main branch to `main`
```suggestion
      - main
```

.github/workflows/dockerhub-description.yml
```diff
@@ -0,0 +1,21 @@
...
+    paths:
+      - DOCKERHUB.md
```
Mogztter, line 6: It can be useful to update the description on hub.docker.com between releases but at the same time we don't want to push changes related to new features before the release.

Maybe we should publish on releases but add `workflow_dispatch` to let us manually trigger an update when needed?

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch

.github/workflows/dockerhub-description.yml
```diff
@@ -0,0 +1,21 @@
...
+      - .github/workflows/dockerhub-description.yml
+jobs:
+  dockerHubDescription:
```
Mogztter, line 9: Please use snake case:
```suggestion
  docker-hub-description:
```

DOCKERHUB.md
```diff
@@ -0,0 +1,66 @@
...
+* [OS X](https://docs.docker.com/docker-for-mac/)
+* [Linux](https://docs.docker.com/get-started/)
+
+### Usage
+
+See also the [installation](https://docs.kroki.io/kroki/setup/configuration/) docs in the [Kroki documentation](https://docs.kroki.io/)
+
+#### start a kroki instance
+
+`$ docker run --name some-kroki -d yuzutech/kroki`
```
Mogztter, line 23: I'm not 100% sure but I think it should be:
```suggestion
    $ docker run --name some-kroki -d yuzutech/kroki
```

DOCKERHUB.md
```diff
@@ -0,0 +1,66 @@
...
+
+This image includes `EXPOSE 8000` (the kroki port), so standard container linking will make it automatically available to the linked containers. If you want to map port 8000 in the container to a port on your host, please use the `--publish` or `-p` flag:
+
+`$ docker run -p8000:8000 --name some-kroki -d yuzutech/kroki`
```
Mogztter, line 27:
```suggestion
    $ docker run -p8000:8000 --name some-kroki -d yuzutech/kroki
```

DOCKERHUB.md
```diff
@@ -0,0 +1,66 @@
...
+
+#### connect with kroki-blockdiag
```
Mogztter, line 29:
```suggestion
#### connect with companion containers
```

DOCKERHUB.md
```diff
@@ -0,0 +1,66 @@
...
+
+We recommend using docker-composer to connect with [kroki-blockdiag](https://hub.docker.com/r/yuzutech/kroki-blockdiag) and/or [kroki-mermaid](https://hub.docker.com/r/yuzutech/kroki-mermaid)
```
Mogztter, line 31:
```suggestion
We recommend using docker-composer to connect with companion containers, such as [kroki-blockdiag](https://hub.docker.com/r/yuzutech/kroki-blockdiag) and/or [kroki-mermaid](https://hub.docker.com/r/yuzutech/kroki-mermaid)
```

DOCKERHUB.md
```diff
@@ -0,0 +1,66 @@
...
+    ports:
+      - "8002:8002"
+```
+
+If you don't want to use `docker-compose`, you can configure the network to connect the two containers and use the `KROKI_BLOCKDIAG_HOST` and `KROKI_BLOCKDIAG_PORT` environment variables to configure the host and port of the kroki-blockdiag server. In the same way, you can use `KROKI_MERMAID_HOST` and `KROKI_MERMAID_PORT` environment variables to configure the host and port of the kroki-mermaid server.
+
+### Environment Variables
+The Kroki image uses several environment variables. While none of the variables are required, they may significantly aid you in using the image.
+
+See the [configuration](https://docs.kroki.io/kroki/setup/configuration/) docs in the [Kroki documentation](https://docs.kroki.io/)
```
Mogztter, line 58: (praise) Good idea to link to the documentation 👍🏻

DOCKERHUB.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Versioning
+
+We use SemVer for versioning.
```
Mogztter, line 62:
```suggestion
We use [SemVer](https://semver.org/) for versioning.
```

DOCKERHUB.md
```diff
@@ -0,0 +1,66 @@
...
+This image includes `EXPOSE 8000` (the kroki port), so standard container linking will make it automatically available to the linked containers. If you want to map port 8000 in the container to a port on your host, please use the `--publish` or `-p` flag:
+
+`$ docker run -p8000:8000 --name some-kroki -d yuzutech/kroki`
+
+#### connect with kroki-blockdiag
+
+We recommend using docker-composer to connect with [kroki-blockdiag](https://hub.docker.com/r/yuzutech/kroki-blockdiag) and/or [kroki-mermaid](https://hub.docker.com/r/yuzutech/kroki-mermaid)
+
+```
+version: "3"
```
Mogztter, line 34: We should probably take this opportunity to add `bpmn` and `excalidraw` companion containers?

DOCKERHUB.md
```diff
@@ -0,0 +1,66 @@
...
+    image: yuzutech/kroki-blockdiag
+    ports:
+      - "8001:8001"
+  mermaid:
+    image: yuzutech/kroki-mermaid
+    ports:
+      - "8002:8002"
+```
+
+If you don't want to use `docker-compose`, you can configure the network to connect the two containers and use the `KROKI_BLOCKDIAG_HOST` and `KROKI_BLOCKDIAG_PORT` environment variables to configure the host and port of the kroki-blockdiag server. In the same way, you can use `KROKI_MERMAID_HOST` and `KROKI_MERMAID_PORT` environment variables to configure the host and port of the kroki-mermaid server.
```
Mogztter, line 53:
```suggestion
If you don't want to use `docker-compose`, you can configure the host and port for each companion containers using environment variables:

|Container|Host|Port|
|--|--|--|
|`kroki-blockdiag`|`KROKI_BLOCKDIAG_HOST`|`KROKI_BLOCKDIAG_PORT`|
|`kroki-mermaid`|`KROKI_MERMAID_HOST`|`KROKI_MERMAID_PORT`|
|`kroki-bpmn`|`KROKI_BPMN_HOST`|`KROKI_BPMN_PORT`|
|`kroki-excalidraw`|`KROKI_EXCALIDRAW_HOST`|`KROKI_EXCALIDRAW_PORT`|
```

.github/workflows/dockerhub-description.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Docker Hub Description
+        uses: peter-evans/dockerhub-description@v2
```
Mogztter, line 15:
```suggestion
        uses: peter-evans/dockerhub-description@v2.4.2
```

.github/workflows/dockerhub-description.yml
```diff
@@ -0,0 +1,21 @@
...
+    branches:
+      - master
+    paths:
+      - DOCKERHUB.md
+      - .github/workflows/dockerhub-description.yml
+jobs:
+  dockerHubDescription:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
Mogztter, line 12:
```suggestion
      - uses: actions/checkout@v2.3.4
```

Mogztter:
Thanks @lind!

> I'm new to GitHub Actions but the action mentioned in the #698 should work. But I don't know if there is a way to test this without actually change the description in Dockerhub.

I don't think so but it's not a big deal since I can easily restore the description if something goes wrong.

> I''m not sure the README is the right file to use as the Dockerhub description since the README includes build information etc. not relevant to Dockerhub users and also there are a need for multiple Dockerhub descriptions, one for each docker image; bmpn, excalidraw...

You are 💯 right, we should use a dedicated file.

> Proposes one description file for each image in the same dir as the Dockerfile /bpmn, /excalidraw...

I agree but let's do that in a follow-up pull request.

> The proposed DOCKERHUB.md is just a copy and slight simplification of the description on Dockerhub today and needs some more work.

I left a few suggestions but overall 👍🏻 

> What do you think @Mogztter

It looks very good, thanks for taking care of this.

.github/workflows/dockerhub-description.yml
lind: Agree that is better.

lind:
> > Proposes one description file for each image in the same dir as the Dockerfile /bpmn, /excalidraw...
> 
> I agree but let's do that in a follow-up pull request.

Good idea. Let's get this to work for the main docker hub repository first.

DOCKERHUB.md
lind: Yes I think it's better to reduce the documentation to a minimum in the docker hub description and if something is missing then add it to the official docs instead. Easier to have updated and consistent docs.

.github/workflows/dockerhub-description.yml
lind: Good idea. I'll try to add both manual and release event.

lind: Due to the [github actions docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-a-list-of-events) i think `[workflow_dispatch, release]:` will work and cover both manual and release events.

DOCKERHUB.md
lind: 👍 Added

DOCKERHUB.md
```diff
@@ -0,0 +1,83 @@
...
+      - "8003:8003"
+  excalidraw:
+    image: yuzutech/kroki-excalidraw
+    ports:
+      - "8004:8004"
+```
+
+If you don't want to use `docker-compose`, you can configure the host and port for each companion container using environment variables:
+
+|Container         | Host                  | Port                  |
```
lind, line 63: I did format the table better. 

(I don't think docker hub description could be written in asciidoc unfortunately)

.github/workflows/dockerhub-description.yml
lind: I think the [Create release](https://github.com/yuzutech/kroki/blob/main/.github/workflows/release.yml#L23) step will trigger the `docker-hub-description` job...

Maybe the `branches:` and `paths:` should be removed then?

Mogztter: I prefer to trigger the workflow when a tag is pushed. The `release` event has types associated: published, unpublished, created, edited, deleted, prereleased, released.
So, as far as I understand, it can be triggered more than once for a release/tag.

Since `workflow_dispatch` does not take any option, I would rather use the "long form" syntax:

```
on:
  workflow_dispatch:
  push:
    tags:
      - 'v*'
```

Mogztter: > And it seems like the create-release action used in the release.yml is archived. Maybe try something like intuit/auto?

I'm not entirely sure why because it was quite useful... and last time I checked alternatives were not as good/simple. So for now, I'm sticking to it 😅

.github/workflows/dockerhub-description.yml
```diff
@@ -0,0 +1,21 @@
+on:
+  [workflow_dispatch, release]:
+    branches:
+      - main
+    paths:
+      - DOCKERHUB.md
```
Mogztter, line 6:
```suggestion
on:
  workflow_dispatch:
  push:
    tags:
      - 'v*'
```

Mogztter:
Thanks @lind, it's working great 🎉 
The description is now up-to-date: https://github.com/yuzutech/kroki/actions/workflows/dockerhub-description.yml

Issue #797, opened 2021-06-16
Update install.adoc with basic podman and podman-compose intro guides


sturtison:
When implementing Kroki with podman and hitting some (non Kroki) issues, I noticed there was no starting point for podman in the docs.

Using podman locally is a also great stepping stone on the journey to Kubernetes.

I've forked this repo and made some small changes to `install.adoc` to introduce using podman with kroki at https://github.com/sturtison/kroki/blob/add-podman-doco-jun15/docs/modules/setup/pages/install.adoc

If it's suitable, then I look for guidance to fit in with the way you work re assessing review, pull requests, squash merges etc.

PR #798, opened 2021-06-16
docs(install.adoc) Add podman guide

sturtison:
This PR addresses issue #797

There are multiple commented commits on the same file.

### DIFF
--- a/docs/modules/setup/pages/install.adoc
+++ b/docs/modules/setup/pages/install.adoc
@@ -13,16 +13,12 @@
 :uri-mermaid: https://mermaidjs.github.io/
 :uri-bpmn: https://bpmn.io/toolkit/bpmn-js/
 :uri-excalidraw: https://excalidraw.com/
-:uri-docker-get-started: https://docs.docker.com/get-started/
-:uri-podman-get-started: https://podman.io/getting-started/
-:uri-docker-compose-install: https://docs.docker.com/compose/install/
-:uri-podman-compose-install: https://github.com/containers/podman-compose
 :uri-docker-container-networking: https://docs.docker.com/config/containers/container-networking/
 :uri-podman-container-networking: https://podman.io/getting-started/network
 :uri-k8s-publish-service: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
 :uri-k8s-definition-files: https://raw.githubusercontent.com/yuzutech/kroki/master/docs/modules/setup/examples/k8s
-:uri-localhost-8000: pass:q[`http://localhost:8000[http://localhost:**8000**]`]
-:uri-localhost-8080: pass:q[`http://localhost:8080[http://localhost:80**80**]`]
+:uri-localhost-8000: pass:q[\http://localhost:**8000**]
+:uri-localhost-8080: pass:q[\http://localhost:**8080**]
 
 To convert plain text diagrams to images with Kroki,
 you first need to install the gateway server.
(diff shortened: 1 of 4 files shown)


docs/modules/setup/pages/install.adoc
```diff
@@ -64,7 +67,14 @@ docker run -p8080:8000 -d yuzutech/kroki
 == Using docker-compose
 
 You can use `docker-compose` to run multiple containers.
-Here's an example where we start all the containers:
+
+Assumptions:
+
+* [x] {uri-podman-get-started}[Podman] is installed on your machine.
+* [x] {uri-podman-compose-install}[Podman-compose] is installed on your machine.
```
Mogztter, line 74: I believe that this section should be under "Using Podman" right?
The assumption in the "docker-compose" is:

```suggestion
Assumptions:

* [x] {uri-docker-compose-install}[Docker Compose] is installed on your machine 
```

docs/modules/setup/pages/install.adoc
```diff
@@ -75,6 +85,44 @@ include::example$kroki-docker-compose.yml[]
...
+
+Podman is a drop-in replacement for Docker for Linux systems only, and is available for many distributions.
+It is light-weight, does not need a daemon running in the background, and is built with container security in mind.
+
+Using Podman is a useful learning, development and testing step for running pods of containers in your linux environment.
+It is also a useful intermediate step if you are looking to lift your Kroki implementation into a Kubernetes environment.
+
+Assumptions:
+
+* [x] {uri-docker-get-started}[Docker] is installed on your machine.
```
Mogztter, line 98:
```suggestion
Assumptions:

* [x] {uri-podman-get-started}[Podman] is installed on your machine.
* [x] {uri-podman-compose-install}[Podman-compose] is installed on your machine.
```

docs/modules/setup/pages/install.adoc
```diff
@@ -64,7 +67,14 @@ docker run -p8080:8000 -d yuzutech/kroki
...
 You can use `docker-compose` to run multiple containers.
-Here's an example where we start all the containers:
+
+Assumptions:
+
+* [x] {uri-podman-get-started}[Podman] is installed on your machine.
+* [x] {uri-podman-compose-install}[Podman-compose] is installed on your machine.
+
+Here's an example where we start all the containers.
+In this example host port 8000 is exposed.
```
Mogztter, line 77: 👍🏻

Mogztter:
@sturtison It looks great, thanks!
I made a few suggestions but I don't want to apply them without your approval. Let me know if they make sense.

sturtison:
Thanks for the suggestions and your time spent reviewing!

Considering them made me look hard at the Kroki install guide once more and think about what it's really trying to do: Get people started by running the Kroki gateway server, then running multiple Kroki containers together (which is what makes Kroki so good).

If I'd continued the way I had, whilst correct, it would have confusingly fragmented your original aesthetic of concise minimal documentation.

So I've taken your suggestions into account while rewriting the whole top section, integrating docker and podman into the tasks above, while hopefully replicating your aesthetic. Should there be the need for a tutorial, that would be a separate document or exercise.

sturtison:
I've made updates based on your suggestions and my reflections and its now suitable for review. Any errors are mine.

Hopefully its inline with what you would like to see in the Kroki docs.

sturtison:
Thank you for the consideration in applying your commits.
What I did do to get to this point, was to work locally (online with a gist) with many commits that you will appreciate not being in this PR.
Please apply or not your suggestions to the commits as you feel most appropriate to keep things moving. It's a good feature of GitHub.

Mogztter:
Thanks for the updates! It's definitely better 👍🏻 
I want to build the documentation locally with your changes to make sure that everything looks good before merging.

I will try to take some time later this week to get it done.

Mogztter:
I made a few changes but overall it was really good 🙌🏻 

I decided to extract "About Docker and Podman for Kroki" in a dedicated page to keep the "Install" page relatively short.

As a follow-up, I think we should extract "Use Kubernetes" in a dedicated page as well. The "Install" menu could look like:

- Install
	- Using Docker or Podman
	- Using Kubernetes
	- Manual Install

Mogztter:
And it's live! 🎉 

https://docs.kroki.io/kroki/setup/install/

sturtison:
Great!

Yes, that change would be good. I could look at that.

Mogztter:
> Yes, that change would be good. I could look at that.

That would be great, thanks again for improving the documentation 👍

PR #808, opened 2021-06-22
correct port for excalidraw k8s service

lind:


Mogztter:
Nice catch and thanks for the quick fix! 👍🏻

PR #857, opened 2021-08-09
resolves #842 creates SVG and PNG error images

### LINKED ISSUE #842: When there's a syntax error and the Accept request HTTP header is "image/*" we should return an image
Currently, Kroki will return the error message as plain text.
As a result, we get a broken image icon when using an `` element:

```html
<html>
<body>
<img src="https://kroki.io/plantuml/svg/eNpLstFVTAQABAkBTg=="/>
</body>
</html>
```

**Chrome**
image

**Firefox**
image

Instead, it would be useful to show the error message as an image:

image

Mogztter:
Use svgSalamander to create SVG and PNG images.
For reference, Batik adds ~7mb of dependencies.

resolves #842

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,6 +3,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.kroki</groupId>
   <artifactId>kroki</artifactId>
+  <name>kroki</name>
   <packaging>pom</packaging>
   <version>0.14.0</version>
   <properties>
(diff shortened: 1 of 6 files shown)


Mogztter:
```html
<img src="https://kroki.io/plantuml/svg/eNpLstFVTAQABAkBTg=="/>
```

And here's the result in a browser (Chrome):

|Before|After|
|--|--|
|image|image|

Issue #867, opened 2021-08-19
Feature request : Please support   public URI as diagram_source value


jvmlet:
Would be great to be able to render diagram with this api:
```
GET /type/format/proxy?diagram_source=https://raw.github.com/myTextDiagram
```

This will enable users to include kroki  generated diagram in Github's   readme adoc/md files. 
Planutml server supports this with `/plantuml/proxy?src=PUBLIC_URL` parameter  (described [here ](https://blog.anoff.io/2018-07-31-diagrams-with-plantuml/) under `GitHub integration section` )

Mogztter:
This is somehow related to: https://github.com/yuzutech/kroki/issues/3
As mentioned in this issue:

> The public instance of Kroki does not do any remote call for performance/security reasons so this approach won't work.
I purposely want to keep the API as simple as possible. It takes a text diagram as input and returns an image.

I guess anyone could work on a proxy to fetch text diagrams from remote URLs and use Kroki as a backend to convert them to images. It could even be an additional services but I don't want to include it in core (at least for now).

marcosvst:
Hey guys!

I'm not sure if it would fit for the suggestions you gave regarding adding a feature to generate the diagram's directly from its source, but I had a similar need for dealing with my own diagrams and I created a tool to help me with that. It's been of a great help so far so I hope it will help someone as well.

It uses Kroki to generate the diagram, but it keeps the source and you can embed to Notion, for example. Check it here https://diagramit.herokuapp.com/.

Thanks!

Mogztter:
Hey @marcosvst, thanks for sharing.

> I'm not sure if it would fit for the suggestions you gave regarding adding a feature to generate the diagram's directly from its source, but I had a similar need for dealing with my own diagrams and I created a tool to help me with that. It's been of a great help so far so I hope it will help someone as well.

This is exactly what I had in mind 👍🏻

marcosvst:
Hey @Mogztter I liked the idea and thanks for being so receptive!

I created the pull request as you suggested here #898. Please let me know if something is off or out of the standards.

Thanks!

PR #898, opened 2021-09-10
Add Diagram it mentioning to docs

marcosvst:
Related to this comment https://github.com/yuzutech/kroki/issues/867#issuecomment-915788485

### DIFF
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -1,6 +1,5 @@
 = Kroki Documentation
 :uri-exoscale: https://www.exoscale.com/
-:uri-diagramit: https://github.com/marcosvst/diagramit
 
 [discrete.tagline]
 === Convert plain text diagrams to images !
@@ -19,6 +18,6 @@ Thanks to {uri-exoscale}[Exoscale], which has kindly offered two servers, we can
 However, some parts of the infrastructure are still host on other providers, so *we are still looking for sponsors to cover all the expenses*.
 +
 If you are interested, please 👋 mailto:hello@kroki.io[contact us].
-
-The public instance of Kroki does not do any remote call for performance/security reasons. Also we want to keep the API as simple as possible, so it takes a text diagram as input and returns an image.
-If you want to generate diagrams directly from a public source file, check this other project named {uri-diagramit}[Diagram it] that was built upon Kroki and allow you to generate the diagram only by providing the diagram's public source file path.
++
+NOTE: The public instance of Kroki does not do any remote call for performance and security reasons.
+We also want to keep the API as simple as possible, so it takes a text diagram as input and returns an image.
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/index.adoc
```diff
@@ -18,3 +19,6 @@ Thanks to {uri-exoscale}[Exoscale], which has kindly offered two servers, we can
 However, some parts of the infrastructure are still host on other providers, so *we are still looking for sponsors to cover all the expenses*.
 +
 If you are interested, please 👋 mailto:hello@kroki.io[contact us].
+
+The public instance of Kroki does not do any remote call for performance/security reasons. Also we want to keep the API as simple as possible, so it takes a text diagram as input and returns an image.
+If you want to generate diagrams directly from a public source file, check this other project named {uri-diagramit}[Diagram it] that was built upon Kroki and allow you to generate the diagram only by providing the diagram's public source file path.
```
Mogztter, line 24: I think it might be better to link to: https://diagramit.herokuapp.com/ no?

Mogztter:
Thanks @marcosvst!
The content is great but I'm not convinced that the home page of the documentation is the right place. I think we should create a new page named "Integration" or "Third Party Tools" with a section about "Diagram it".

We could also use this new page to talk about other projects that are related to Kroki (or that uses Kroki).

What do you think?

marcosvst:
> Thanks @marcosvst!
> The content is great but I'm not convinced that the home page of the documentation is the right place. I think we should create a new page named "Integration" or "Third Party Tools" with a section about "Diagram it".
> 
> We could also use this new page to talk about other projects that are related to Kroki (or that uses Kroki).
> 
> What do you think?

You're right, that makes more sense.

Thanks for updating the PR.

Mogztter:
No worries, thanks for reviewing.

Mogztter:
@marcosvst For reference, I get an "Internal Server Error" when using the following URL: https://diagramit.herokuapp.com/v1/plantuml/generate?source=https://raw.githubusercontent.com/yuzutech/kroki/main/tests/smoke/diagrams/architecture.puml

marcosvst:
> @marcosvst For reference, I get an "Internal Server Error" when using the following URL: https://diagramit.herokuapp.com/v1/plantuml/generate?source=https://raw.githubusercontent.com/yuzutech/kroki/main/tests/smoke/diagrams/architecture.puml
> 
> I also noticed that the dropdown only has an handful of options but if you type the name of the diagram library in the URL it's working as expected.
> For instance, Vega is not in the dropdown but you can use:
> 
> https://diagramit.herokuapp.com/v1/vega/generate?source=https://github.com/yuzutech/kroki/blob/main/tests/smoke/diagrams/bar-chart.vega

@Mogztter thanks for spotting that, it was a silly bug but that is fixed now.

Re: dropdown only having a few options - that was expected, the backend is prepared to handle all diagram types Kroki supports, I didn't add them all to the generator webpage yet because of lack of time.

Mogztter:
> Re: dropdown only having a few options - that was expected, the backend is prepared to handle all diagram types Kroki supports

👍🏻 

> I didn't add them all to the generator webpage yet because of lack of time.

No worries, I can relate 😉

PR #904, opened 2021-09-14
Explain how to pass options using the Kroki API

Mogztter:


Mogztter:
@sturtison All suggestions are most welcome 😉

PR #909, opened 2021-09-19
Make sure that yuzutech/kroki image is running fine with Podman

Mogztter:


Mogztter:
Tests completed, it seems to be working fine, closing.

PR #930, opened 2021-10-02
resolves #929 upgrade Structurizr to version 1.15.0

### LINKED ISSUE #929: Upgrade Structurizr DSL to version 1.15.0 and replace structurizr-plantuml in favour of structurizr-export
Follow-up of https://github.com/yuzutech/kroki/issues/510

Mogztter:
resolves #929


server/src/main/java/io/kroki/server/service/Structurizr.java
```diff
@@ -89,10 +97,23 @@ public void convert(String sourceDecoded, String serviceName, FileFormat fileFor
...
+        diagram = structurizrPlantUMLExporter.export((ComponentView) selectedView);
+      } else if (selectedView instanceof ContainerView) {
+        diagram = structurizrPlantUMLExporter.export((ContainerView) selectedView);
+      } else if (selectedView instanceof SystemContextView) {
+        diagram = structurizrPlantUMLExporter.export((SystemContextView) selectedView);
+      } else if (selectedView instanceof SystemLandscapeView) {
+        diagram = structurizrPlantUMLExporter.export((SystemLandscapeView) selectedView);
+      } else {
+        throw new BadRequestException("View type is not supported: " + selectedView.getClass().getSimpleName() + ", must be a DynamicView, DeploymentView, ComponentView, ContainerView, SystemContextView or SystemLandscapeView.");
+      }
```
Mogztter, line 115: @simonbrowndotje Thanks for your guidance, I've replaced structurizr-plantuml in favour of structurizr-export. Not sure if this is the best way to export a `View` to a `Diagram`?

Alternatively, I could export the whole `Workspace` and then filter by view id but I suppose it will be less performant.

Issue #677, opened 2021-04-04
Support HTTP HEAD requests


DaveJarvis:
# Replicate

Perform an HTTP HEAD request as follows:

1. Install httpie
1. Run: `http HEAD https://kroki.io/graphviz/svg/eNpLyUwvSizIUHBRqOZSUEhU0LVTSLLmqgUAVQQGWQ== --verbose`

# Expected

Content type set to `image/svg+xml`.

# Actual

Content type set to `text/plain`.

``` plain
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: kroki.io
User-Agent: HTTPie/2.4.0
HTTP/1.1 404 OK
CF-Cache-Status: DYNAMIC
CF-RAY: 63ad15d1ae0cc520-ORD
Connection: keep-alive
Content-Encoding: gzip
Content-Type: text/plain
Date: Sun, 04 Apr 2021 19:53:52 GMT
Expect-CT: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
NEL: {"report_to":"cf-nel","max_age":604800}
Report-To: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report?s=Yj37MVYec5YKZXI5B7He1vNNsQbhHSDQzOMadNdzlb4aqzcyCviEq%2B5hdxs50GtUYNYTLstBnAQm4rZQohQvB7r5ncmBQLkKLQ%3D%3D"}],"group":"cf-nel","max_age":604800}
Server: cloudflare
Set-Cookie: __cfduid=d996b29d5b1f566da2c82cd014df2d6281617566031; expires=Tue, 04-May-21 19:53:51 GMT; path=/; domain=.kroki.io; HttpOnly; SameSite=Lax; Secure
alt-svc: h3-27=":443"; ma=86400, h3-28=":443"; ma=86400, h3-29=":443"; ma=86400
cf-request-id: 09400bf7040000c5209b012000000001
```

# Rationale

HTTP HEAD requests can reduce bandwidth.

# Return Values

> We will probably need to determine what Kroki should do when it receives a HEAD request.
> For instance, should we try to decode the payload and raise an error if the content is not properly encoded?
> Should we return a 200 on https://kroki.io/graphviz/svg? etc...

[...]

PR #948, opened 2021-10-09
added head http method

amirabramovich:
resolved issue #677

### DIFF
--- a/server/src/main/java/io/kroki/server/service/DiagramRegistry.java
+++ b/server/src/main/java/io/kroki/server/service/DiagramRegistry.java
@@ -34,6 +34,9 @@ public class DiagramRegistry {
         .handler(bodyHandler)
         .handler(diagramHandler.createRequestReceived(name))
         .handler(diagramHandler.createPost(name));
+      router.head("/" + name + "/:output_format/:source_encoded")
+        .handler(diagramHandler.createRequestReceived(name))
+        .handler(diagramHandler.createGet(name));
     }
   }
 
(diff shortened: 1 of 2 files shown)


Mogztter:
Hey, thanks for your contribution!

Before jumping into the implementation, I think we should thoroughly describe how we are going to implement this feature on the GitHub issue.
They are a few open questions and since it will be part of the public API we have to be considerate.

For instance, the HEAD request should not have a body. It should return the `Content-Type` and the `Content-Length` (of the generated diagram) but it should not actually send the diagram.

In this regard, the current implementation is incorrect because it uses `diagramHandler.createGet(name)` which will return the same result whether we use `GET` or `HEAD`.

Anyway, if you want to implement this feature, please use the GitHub issue to describe how you will implement it so we can discuss it.

Mogztter:
> I've opened a feature request https://github.com/eclipse-vertx/vert.x/issues/4150 but in the meantime we should define this header on HEAD requests (using buffer.length()).

Since my request has been rejected on Vert.x, could you please set the `Content-Length` otherwise the caller won't be able to check the file size.

Since we are working on HEAD requests, we might want to return a "200" when calling a valid output format for a given diagram library:

```
$ http HEAD :8000/graphviz/svg
HTTP/1.1 200 OK
```
```
$ http HEAD :8000/graphviz/pdf
HTTP/1.1 404 Not Found
```

Arguably, `GET /graphviz/svg` could return a `204 No Content` or a `400 Bad Request` instead of a 404.

What do you think?

PR #951, opened 2021-10-13
resolves #507 add png support for Mermaid

### LINKED ISSUE #507: Mermaid PNG support
It would be awesome if there were PNG support for Mermaid charts. SVG is not as usable from an API point of view.

derlin:
Add the `/png` endpoint support for mermaid diagrams.
The resulting png has transparent background, and has no margin (will really only contain the actual rendered diagram).

See #507


derlin:
@Mogztter thank you for the review ! 

I applied the proposed changes, fixed the linter issues and added the missing `return` statements after `micro.send` (would make the server crash on missing body otherwise).

I also included two tests about PNG generation using *pixelmatch* as you proposed.  I just had a doubt on where to put the reference images. They are currently under `tests`, but maybe a subfolder (e.g. `tests/resources` would be cleaner ?

Mogztter:
> I applied the proposed changes, fixed the linter issues and added the missing return statements after micro.send (would make the server crash on missing body otherwise).

👍🏻 

> I also included two tests about PNG generation using pixelmatch as you proposed. I just had a doubt on where to put the reference images. They are currently under tests, but maybe a subfolder (e.g. tests/resources would be cleaner ?

That's a good idea.
Usually, I'm using the word "fixtures" but I don't have a strong opinion on the naming.

Mogztter:
Tests are currently failing because the width and height of the PNG images are slightly different. My guess is that rasterization is slightly different (most likely on fonts) and you can get a few pixels difference.


```
  1) #convert
       should return a PNG equal to: graph.png:

      AssertionError: expected 178 to equal 175
      + expected - actual

      -178
      +175
      
      at Context.<anonymous> (tests/test.js:66:26)
      at processTicksAndRejections (internal/process/task_queues.js:95:5)

  2) #convert
       should return a PNG equal to: seq.png:

      AssertionError: expected 504 to equal 500
      + expected - actual

      -504
      +500
      
      at Context.<anonymous> (tests/test.js:66:26)
      at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

So we should also use a threshold on the width and height. Something like, more or less 5% of the expected width/height:

```js
// threshold, more or less 5% of the expected widht/height
expect(width).to.be.within(expectedImage.width / 1.05, expectedImage.width * 1.05);
expect(height).to.be.within(expectedImage.height / 1.05, expectedImage.height * 1.05);
```

derlin:
> Tests are currently failing because the width and height of the PNG images are slightly different. My guess is that rasterization is slightly different (most likely on fonts) and you can get a few pixels difference.

Wow, it passed locally and I didn't think of this. Will add the threshold tomorrow.
Just to ensure I understood, you mean that depending on the os, puppeteer will generate slightly different images, due mostly to the font rendering?

derlin:
> Are you using macOS?

Yes I am. I thought also of simply making the tests run on docker. However, to me having a test that fails locally but succeeds on docker build is more problematic than not. I imagine the next developer cloning the repo, and spending 3h understanding why the tests suddenly fail, because he didn't think of trying to run it inside docker.

But anyways, as you said, first things first: let's see if pixelmatch can handle different sizes (will work on it after work). Will keep you posted!

derlin:
Ok, so I can actually make the code return the error SVG as image, but since this error as SVG doesn't provide any information (except that a problem has occurred), to me it would make sense to instead return `400: BAD REQUEST`, with a body `Syntax Error in Graph`. This way, if you use it from the command-line for example, you can directly know there is a problem without having to actually open the output file (svg or png).

This is actually what most other endpoints are doing: PlantUML, BlockDiag, PacketDiag, etc. 
If this is ok with you, I will change the behavior to `400: BAD REQUEST` for both endpoints, and add a test about it. Would this be ok ?

PR #955, opened 2021-10-16
resolves #787 add support for SSL

### LINKED ISSUE #787: How to run as https with Docker Kroki
I am a user of Docker version of Kroki.

I run the service as the example command: `docker run -p8080:8000 -d yuzutech/kroki`.

It works in http, but now I want to switch to https with my own SSL certificate files.

Cloud please tell me how to do this?

I had tried to read the official documents, but not luck.

Thanks.

amirabramovich:
resolved issue #787

### DIFF
--- a/server/src/main/java/io/kroki/server/Server.java
+++ b/server/src/main/java/io/kroki/server/Server.java
@@ -163,8 +163,10 @@ public class Server extends AbstractVerticle {
         throw new IllegalArgumentException("KROKI_SSL_KEY and KROKI_SSL_CERT must be configured when SSL is enabled.");
       }
       serverOptions.setPemKeyCertOptions(
-        new PemKeyCertOptions().addKeyValue(Buffer.buffer(sslKeyValue.get()))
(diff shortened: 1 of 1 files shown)


Mogztter:
Looks good, left a few suggestions (style) and an open question.
Would you be able to write the documentation for this feature? I guess you can add a section in https://github.com/yuzutech/kroki/blob/main/docs/modules/setup/pages/configuration.adoc

I can tell you that the search term "ssl" and "https" are pretty popular on https://docs.kroki.io/ so it's definitely something users will be interested to read 😄

amirabramovich:
Thanks for your suggestions, I made all the adjustments and added a documentation :)

Mogztter:
> I made all the adjustments and added a documentation :)

Thanks!

I just noticed that we are using `setKeyPath` and `setCertPath` however I think it's easier to pass the contents of the private key and the certificate.

In my experience it makes things easier since you don't have a to mount a volume to provide the certificate and private key files. You can just pass the contents of the certificate and the private key as string using environment variables.

Taking this into account, I've made several suggestions.

PR #972, opened 2021-11-03
doc: the tools rely on Kroki, not the list

cmorty:


Mogztter:
You're probably right, thanks 👍🏻

PR #974, opened 2021-11-03
Use Python-version agnostic helpers

jamietanna:
As it may not be the default `python` on an OS any more, we should make
it possible to run against Python 2 and Python 3.

### DIFF
--- a/server/README.adoc
+++ b/server/README.adoc
@@ -90,7 +90,7 @@ For example:
 $ mkdir example-lib
 $ printf "@startuml\nBob->Alice\n@enduml\n" > example-lib/bob.puml
 $ sudo docker run --publish=8000:8000 -v $(pwd)/example-lib:/example-lib -e KROKI_SAFE_MODE=SAFE -e KROKI_PLANTUML_INCLUDE_PATH=/example-lib yuzutech/kroki:latest
-$ curl http://localhost:8000/plantuml/svg/$(printf '@startuml\n!include bob.puml\nAlice -> Bob\n@enduml' | python -c "import sys; import base64; import zlib; print(base64.urlsafe_b64encode(zlib.compress(sys.stdin.read(), 9)))")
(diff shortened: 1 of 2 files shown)


jamietanna:
Amended with suggested fixes from https://github.com/yuzutech/kroki.io/pull/80

server/README.adoc
```diff
@@ -90,7 +90,7 @@ For example:
 $ mkdir example-lib
 $ printf "@startuml\nBob->Alice\n@enduml\n" > example-lib/bob.puml
 $ sudo docker run --publish=8000:8000 -v $(pwd)/example-lib:/example-lib -e KROKI_SAFE_MODE=SAFE -e KROKI_PLANTUML_INCLUDE_PATH=/example-lib yuzutech/kroki:latest
-$ curl http://localhost:8000/plantuml/svg/$(printf '@startuml\n!include bob.puml\nAlice -> Bob\n@enduml' | python -c "import sys; import base64; import zlib; print(base64.urlsafe_b64encode(zlib.compress(sys.stdin.read(), 9)))")
+$ curl https://kroki.io/plantuml/svg/$(printf '@startuml\n!include bob.puml\nAlice -> Bob\n@enduml' | python -c "import sys; import base64; import zlib; print(base64.urlsafe_b64encode(zlib.compress(sys.stdin.read().encode('utf-8'), 9)).decode('ascii'))")
```
Mogztter, line 93: In this example, we are running the server locally with `docker`, so the URL should remain http://localhost:8000

```suggestion
$ curl http://localhost:8000/plantuml/svg/$(printf '@startuml\n!include bob.puml\nAlice -> Bob\n@enduml' | python -c "import sys; import base64; import zlib; print(base64.urlsafe_b64encode(zlib.compress(sys.stdin.read().encode('utf-8'), 9)).decode('ascii'))")
```

jamietanna:
Sorry @Mogztter, good spot, that was a bad copy-paste!

Mogztter:
> Sorry @Mogztter, good spot, that was a bad copy-paste!

No worries, all good now, thanks 👍🏻

PR #1065, opened 2022-01-14
resolves #1064 Fixes date-related HTTP headers

### LINKED ISSUE #1064: HTTP date-related headers wrong
The date-related headers seen by the web browser (through an Nginx proxy) are as follows:
```http
Date: Fri, 14 Jan 2022 21:50:44 GMT
Expires: 1642629044863
Last-Modified: 1567581178724
```
* `Expires` and `Last-Modified` are milliseconds since the Epoch
* `Last-Modified` is the "`compileTime`" (see #1062)
* The `Date` was probably fixed by Nginx, because it is also generated as the number of milliseconds by `Caching.java`

https://github.com/yuzutech/kroki/blob/c4e0bb386f4742bd788d5fa6b53c2d49100f9e4c/server/src/main/java/io/kroki/server/response/Caching.java#L27

I'll be preparing a pull request, but can only really test it after #1063.

MarcelWaldvogel:
Changes date headers to the HTTP format.

(Not tested yet, please wait with merging)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,5 +1,4 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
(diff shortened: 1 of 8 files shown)


MarcelWaldvogel:
Works as it should. Generating a test for it seems overly complicated, so I'll refrain from it for now. I hope you agree.

Mogztter:
>  Generating a test for it seems overly complicated, so I'll refrain from it for now. I hope you agree.

It should be relatively straightforward if `today` is an optional parameter (default value is `System.currentTimeMillis()`). Then, in your test you can use a fixed timestamp.

But first let's agree on what we should do (in the issue)

Mogztter:
Sorry, I just saw that you pushed new commits, I will review 👍🏻

Mogztter:
Made a few adjustments but otherwise it's great, thanks 🎉

PR #1066, opened 2022-01-15
Shrink docker build context

MarcelWaldvogel:
During `make buildDockerImages`, I noticed that 1.8GB (or GO :wink:) have to be sent to the docker builder. This significantly reduces the context that has to be transferred.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.git/
+*.iml
+.vertx/
+**/node_modules/


Mogztter:
Oh that's why, I didn't know that!
I found a good explanation:

> The Docker client sends the entire "build context" to the Docker daemon. That build context (by default) is the entire directory the Dockerfile is in (...)

We should probably add a `.dockerignore` in each directory from where we build a Docker image?

And we might reduce the build context by building from a subdirectory when generating binaries with Docker.

Anyway, I'm glad to merge this pull request as is since it already improves the situation. Let me know if you think we should add additional improvements in this pull request or in follow-up pull requests?

MarcelWaldvogel:
[Only the `.dockerignore` at the root of the current context is consulted](https://docs.docker.com/engine/reference/builder/#dockerignore-file), so, yes, for different contexts, different `.dockerfile`s are needed.

As you are creating many different contexts, limiting the inclusion of binaries might be useful. But it is also possible to create multiple images in a single `docker build` process (multiple [`FROM` lines](https://docs.docker.com/engine/reference/builder/#from) in a single `Dockerfile`). AFAIK, however, only one of the multiple builds, namely the last, can be easily made into a "public" image.

However, if there are just a few dozen megabytes sent to the Docker context, I typically do not care. Only what is used with `COPY` and friends ends up in the final image, everything else is only slowing down build time (and not that much, compared to the rest). But once I see the transfer counter several seconds, I try to reduce the context size…

MarcelWaldvogel:
BTW: If you think building for ARM might be a good idea, I do have experience creating [multi-architecture docker images](https://netfuture.ch/2020/05/multi-arch-docker-image-easy/), e.g. in [KNOT DNS](https://gitlab.nic.cz/knot/knot-dns/-/blob/master/scripts/Makefile.docker#L23-26).

Mogztter:
> However, if there are just a few dozen megabytes sent to the Docker context, I typically do not care. Only what is used with COPY and friends ends up in the final image, everything else is only slowing down build time (and not that much, compared to the rest). But once I see the transfer counter several seconds, I try to reduce the context size…

That sounds fair. Let's merge this pull request 👍🏻 

> BTW: If you think building for ARM might be a good idea, I do have experience creating multi-architecture docker images, e.g. in KNOT DNS.

There might be some interest, please create a new issue so we can evaluate if we want to support it or not.

Issue #1037, opened 2021-12-22
Very high CPU with v0.16.0


max-wittig:
Seems like that since upgrading to version 0.16.0, the machine requirements increased a lot. Before our machine was almost idle at maximum ~25%. Now it's always going to 100%.

Is this expected?

<img width="683" alt="image" src="https://user-images.githubusercontent.com/6639323/147075556-a2e249af-9d9d-4bfd-8cbb-d67da9f4c886.png">

<img width="1428" alt="image" src="https://user-images.githubusercontent.com/6639323/147075592-7a8c0c1f-6320-4030-b0f8-c91d3941e2ad.png">

@Mogztter

/cc @dlouzan @ercanucan

Mogztter:
I recently discovered an infinite loop in PlantUML [1] which is already fixed on the latest snapshot but in your case it seems that Chrome is taking a lot of CPU.
I've seen some issues with Mermaid in the past but it would interesting to find out which diagram library and which diagram in particular is causing this issue. You can use `docker logs` to get the logs from `yuzutech/kroki` between 10:00 and 10:30.

If it's indeed related to a particular container, as a workaround, you can restart it.

[1] https://github.com/plantuml/plantuml/issues/836

max-wittig:
Not sure. We have a lot of traffic, but it's almost completely `plantuml` actually. With some `ditaa` in between

max-wittig:
We've just increased the machine size now to avoid issues over the holidays. But looks like maybe it was just a temporary thing.

Mogztter:
Thats odd... PlantUML and ditaa are not using Chrome.
Maybe a Chrome instance went haywire during the upgrade/restart of the bpmn, mermaid or excalidraw container.

Anyway, if you find the root cause or investigate further please share the results (even in raw form).

hof:
I'm seeing the same thing. After starting the latest containers immediately seeing a bunch of `chrome` processes in `mermaid`, `bpmn` and `excalidraw` that take the CPU to 100%. 

The logs of `mermaid`, `bpmn` and `excalidraw` just show the endpoint only. 

The `kroki` container shows a warning: 

```
{"timestamp":"1640914105494","level":"WARN","thread":"vertx-blocked-thread-checker","logger":"io.vertx.core.impl.BlockedThreadChecker","message":"Thread Thread[vert.x-eventloop-thread-1,5,main] has been blocked for 2294 ms, time limit is 2000 ms","context":"default"}
[...]

PR #1166, opened 2022-02-25
ref #1037 use a single process (to avoid deadlock and high CPU usage)

Mogztter:


Mogztter:
Smoke tests were stalled on `excalidraw/svg should answer with HTTP 200`, might indicate an issue, merge with caution!

PR #1178, opened 2022-03-06
doc: use expose instead of ports in docker-compose.yml

arukiidou:
following up at https://github.com/yuzutech/kroki/pull/149


Mogztter:
Thanks!

PR #1179, opened 2022-03-06
chore(smoketests) don't expose companion container ports

arukiidou:
following up at https://github.com/yuzutech/kroki/pull/149

and https://github.com/yuzutech/kroki/pull/1178

The difference is that, the above is a "free" fix for the document only, 

this is fix for the test, so it's possible that it could break the test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,10 +72,6 @@ smokeTests:
 	&& docker-compose --file "$(SMOKE_TESTS_DIR)/docker-compose.yaml" ps \
 	&& echo \
 	&& "$(SMOKE_TESTS_DIR)/wait-for-it.sh" localhost:8000 --timeout="$(COMPOSE_TIMEOUT)" \
-	&& "$(SMOKE_TESTS_DIR)/wait-for-it.sh" localhost:8001 --timeout="$(COMPOSE_TIMEOUT)" \
-	&& "$(SMOKE_TESTS_DIR)/wait-for-it.sh" localhost:8002 --timeout="$(COMPOSE_TIMEOUT)" \
-	&& "$(SMOKE_TESTS_DIR)/wait-for-it.sh" localhost:8003 --timeout="$(COMPOSE_TIMEOUT)" \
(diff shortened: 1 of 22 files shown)


Mogztter:
@arukiidou Could you please rebase? I just merged #1178

Mogztter:
@arukiidou smoke tests are failing because we are using `wait-for-it.sh` on port 8001, 8002, 8003 and 8004 in order to make sure that all companion containers are up and running before executing the tests suite.

https://github.com/yuzutech/kroki/blob/eabc978298641a5044b3ec85068649c139cdbacd/Makefile#L74-L78

Since ports 8001, 8002, 8003 and 8004 are not exposed on the host we will need another strategy to make sure that the companion containers are up and running.

Maybe using `wait-for-it.sh` on port 8000 and waiting 15 seconds is enough. You could try to remove lines 75 to 78 and see what happens.

arukiidou:
> @arukiidou smoke tests are failing because we are using `wait-for-it.sh` on port 8001, 8002, 8003 and 8004 in order to make sure that all companion containers are up and running before executing the tests suite.
> 
> https://github.com/yuzutech/kroki/blob/eabc978298641a5044b3ec85068649c139cdbacd/Makefile#L74-L78
> 
> Since ports 8001, 8002, 8003 and 8004 are not exposed on the host we will need another strategy to make sure that the companion containers are up and running.
> 
> Maybe using `wait-for-it.sh` on port 8000 and waiting 15 seconds is enough. You could try to remove lines 75 to 78 and see what happens.

Success.

You have two choices. Which is better? 
- close this pr. because the smoke test are legitimate reasons to use ports, not expose. 
- Instead, test it using

```
>docker compose ps | grep -e blockdiag | grep -e running  | grep -e  8001/tcp
blockdiag           "gunicorn --preload …"   blockdiag           running             8001/tcp
>WAITFORIT_result=$?
>echo $WAITFORIT_result
0
```

Mogztter:
Let's keep it that way, we can always revisit it if we need/want something more sophisticated.
Thanks!

PR #1227, opened 2022-03-20
Build multi-arch (arm64 + amd64) images for excalidraw, mermaid, bpmn and diagramsnet

Mogztter:


Mogztter:
It seems to be working, now I need to find a way to integrate `buildx` in the build/release process.

PR #1275, opened 2022-05-26
Preserve classy stdlib includes

anb0s:


anb0s:
classy-c4 has an [issue](https://github.com/plantuml/plantuml-stdlib/issues/56) and cannot be used -> will add only classy

Mogztter:
@anb0s could you please rebase on main?

PR #1296, opened 2022-07-12
Align 'unsupported output format' error messages

bauglir:
Updates the "unsupported output format" error message for non-HTML responses to be aligned with the HTML response.

The previous implementation caused the `serviceName` to be used in place of the `supportedFormat` and vice versa, e.g. `Unsupported output format: jpg. Must be one of mermaid for png or svg.`

The other option would have been to swap the arguments, but I figured the consistent error message was preferred.


Mogztter:
Nice catch and thanks for the fix 😊

PR #1310, opened 2022-08-17
docs: Add reference to MkDocs Kroki plugin

fviolette:
Listing Hugo and MkDocs integrations.

For Hugo, found multiple community sources, but I could not say which one works (best):

https://github.com/netspective-studios/hugo-shortcode-diagram
https://github.com/jmooring/hugo-testing/tree/hugo-forum-topic-36924

Can this be integrated nonetheless?

### DIFF
--- a/docs/modules/setup/pages/third-party-tools.adoc
+++ b/docs/modules/setup/pages/third-party-tools.adoc
@@ -15,6 +15,8 @@
 :url-julia-int: https://bauglir.github.io/Kroki.jl/stable
 :url-julia-pluto: https://github.com/fonsp/Pluto.jl
 :url-julia-vscode: https://www.julia-vscode.org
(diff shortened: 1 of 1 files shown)


Mogztter:
Hey François, thanks for these additions!
I don't mind adding MkDocs integrations but I think it's a bit too soon for Hugo. 

- https://github.com/netspective-studios/hugo-shortcode-diagram has very few commits, low activity and only one start. It also runs client-side which is, in my opinion, suboptimal.
- https://github.com/jmooring/hugo-testing/tree/hugo-forum-topic-36924 has ten stars but the installation is a bit cumbersome

Kroki has an "official" Go library (https://github.com/yuzutech/kroki-go) and since Hugo is written in Go I believe there might be better ways to integrate Kroki in a Hugo site. (I'm not familiar with Hugo but I will take a closer look).

fviolette:
I'll remove the Hugo part for now and try and reach out for a better, more stable source.

PR #1318, opened 2022-08-26
resolves #970 use cargo.toml to install svgbob

### LINKED ISSUE #970: Use a Cargo.toml file to build the static binary of svgbob
Currently, the version of svgbob is declared in a Dockerfile: https://github.com/yuzutech/kroki/blob/main/server/ops/docker/build-static-svgbob

As a result, the version is not automatically updated by Dependabot.

ssaraswati:
Addresses #970 by converting the build-static-svgbob to use a Cargo.toml file instead of specifying version directly

I wasn't able to get a full build of kroki to run on my machine but this did seem to work correctly with `COPY --from=kroki-builder-static-svgbobs:latest /home/rust/.cargo/bin/svgbob /usr/bin/svgbob` from `server/ops/docker/jdk11-alpine/Dockerfile`

### DIFF
--- a/server/ops/docker/Cargo.toml
+++ b/server/ops/docker/Cargo.toml
@@ -1,0 +1,4 @@
+# Specify svgbob_cli via Cargo.toml so that Renovate can help with updates
+
+[dependencies]
+svgbob_cli = "0.5.3"
--- a/server/ops/docker/build-static-svgbob
+++ b/server/ops/docker/build-static-svgbob
@@ -1,4 +1,7 @@
 # build static executable binary
 FROM ekidd/rust-musl-builder:1.56.1
+COPY Cargo.toml .
 
-RUN cargo install --quiet --version 0.5.3 svgbob_cli
+RUN SVGBOB_VERSION=`cat Cargo.toml | grep "svgbob_cli =" | sed -r 's/.*"([^"]+)"/\1/'` \
+  && cargo install --quiet --version $SVGBOB_VERSION svgbob_cli
+


server/ops/docker/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+# Specify svgbob_cli via Cargo.toml so that Dependabot can help with updates
```
Mogztter, line 1: Actually, we are now using renovate

server/ops/docker/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[package]
+name = "kroki_svgbob_cli"
+version = "0.1.0"
+edition = "2021"
```
Mogztter, line 6: These attributes are mandatory but won't be used, is that correct?

server/ops/docker/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[[bin]]
+name = "svgbob_cli"
+path = "/home/rust/.cargo/bin/svgbob"
```
Mogztter, line 10: I don't understand why we need to specify a `path` since the binary ends up in `/home/rust/.cargo/registry/index/*/.cache/sv/gb/svgbob`?

Mogztter:
Thanks for taking care of this issue 👍🏻 
I left a few comments but it looks very good.

> I wasn't able to get a full build of kroki to run on my machine

Let's see what the CI says.

ssaraswati:
> Thanks for taking care of this issue 👍🏻 I left a few comments but it looks very good.
> 
> > I wasn't able to get a full build of kroki to run on my machine
> 
> Let's see what the CI says.

HI @Mogztter from the failed CI run it doesn't look like my initial approach is likely to work. I have taken a second go doing this via SED to extract the svgbob_cli version from the Cargo.toml file instead.

This approach seems to work fine with Renovate and doesn't require the full Cargo.toml file values to be specified. I have run a copy of renovate against it on my fork and PR was created with new svgbob versions https://github.com/ssaraswati/kroki/pull/3/files

server/ops/docker/build-static-svgbob
```diff
@@ -1,4 +1,7 @@
 # build static executable binary
 FROM ekidd/rust-musl-builder:1.56.1
+COPY Cargo.toml .
+
+RUN SVGBOB_VERSION=`cat Cargo.toml | sed -nr '/svgbob_cli =/{s/svgbob_cli = "([0-9]+.[0-9]+.[0-9]+)"/\1/;p;}'` \
```
Mogztter, line 5: Maybe we can simplify the command using grep?

```
cat Cargo.toml | grep "svgbob_cli =" | sed -r 's/.*"([^"]+)"/\1/'
```

I don't think it needs to check the value enclosed in double quotes.

What do you think?

ssaraswati: Works for me, have updated it

PR #1352, opened 2022-10-24
Add an Elixir encode example

zorbash:
Enhances the docs -> https://docs.kroki.io/kroki/setup/encode-diagram to include an Elixir example encoding a graph.


Mogztter:
Perfect, thank you 🤗

PR #1360, opened 2022-11-01
Build using GitHub cache

Mogztter:


Mogztter:
That's quite nice: "CI / test-containers (pull_request) Successful in 4m" ⚡ 
Usually, it takes a little less than 30minutes.

Mogztter:
With multi-arch (amd64/arm64) it takes 21m but it's still a nice improvement.

PR #1367, opened 2022-11-07
Test that GraphViz supports jpeg output

Mogztter:


Mogztter:
A new version of GraphViz was pushed to Alpine 3.16 🎉

PR #1406, opened 2022-12-11
Added examples for K8s single-pod Deployment

WilliamMiceli:
Found that using K8s, you can vastly simplify configuration very easily, thanks to the containers using different ports by default.

If you want to keep, not sure how you'd like to include it in your docs, but can try to help with that if you'd like.


WilliamMiceli:
Alrighty, I believe I got it all. How does it look to you?

WilliamMiceli:
I shouldn't have merged, haha. Maybe this time it'll work :P

docs/modules/setup/examples/k8s-single-pod/kroki-deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+    matchLabels:
+      app: kroki
+  template:
+    metadata:
+      labels:
+        app: kroki
+    spec:
+      containers:
+      - name: kroki
+        image: docker.io/yuzutech/kroki:0.18.0
```
Mogztter, line 20: It's a good practice to set a version but I think we should use "latest" or update this version when we release a new version. 
My concern is that someone in a few months/years will copy/paste this deployment and run an outdated version.

Mogztter:
All good now! I left a comment about the version in the k8s configuration. I think, for now, we should use "latest". In a second step, I will update the release script to set the actual latest version and automatically create a zip with the updated files.

docs/modules/setup/examples/k8s-single-pod/kroki-deployment.yaml
WilliamMiceli: Very true, wasn't thinking about that, mainly just copied & pasted from what I was using in my own cluster, haha.

Mogztter:
Thank you very much!

> No guarantees, but I might be able to figure out how to do the "second step" you mentioned. I rarely mess with anything CI, but I'll see if I can give it a shot, haha.

No worries.
Currently, I'm running `RELEASE_VERSION=1.2.3 make setServerVersion` to set the version with Maven but we could expand this tasks (and rename it `release`).

https://github.com/yuzutech/kroki/blob/0139dda94b719f6931731bdedcbae726d37a4637/Makefile#L16-L20

SayakMukhopadhyay:
Hey there @Mogztter. It seems that this is not yet reflected in the docs. I see that the job https://github.com/yuzutech/docs.kroki.io/actions/runs/3688866657/jobs/6244166461 was triggered and completed successfully but I still see the older stuff
image
Moreover, I am able to download `kroki-k8s.zip` but not `kroki-k8s-multipod.zip` or the single pod version from the docs site.

Did something break in the CI??

Mogztter:
Thanks for the heads-up!
It looks like a cache issue on Netlify, I've rebuilt the site without cache and now I think we are up-to-date.

PR #1415, opened 2023-01-01
Add Chromium flags for arm64

Mogztter:


mermaid/src/browser-instance.js
```diff
@@ -5,12 +5,58 @@ const { logger } = require('./logger')
 const createBrowser = async () => {
   const browser = await puppeteer.launch({
     dumpio: true,
+    // reference: https://peter.sh/experiments/chromium-command-line-switches/
     args: [
```
Mogztter, line 9: #### TODO

- [ ] remove duplicated flags: https://github.com/puppeteer/puppeteer/blob/6bc152ece01371da519bc942967540751fd93b2a/packages/puppeteer-core/src/node/ChromeLauncher.ts#L179-L212 (default args)

PR #1437, opened 2023-01-19
Add support for PDF for Mermaid

tvh:


Mogztter:
Thanks!
Could you please add a smoke test in: https://github.com/yuzutech/kroki/blob/a127cc2b29567db61713db8ecddc587a02b2a2b8/ci/tests/smoke.js#L18

We also need to allow PDF format in: https://github.com/yuzutech/kroki/blob/a127cc2b29567db61713db8ecddc587a02b2a2b8/server/src/main/java/io/kroki/server/service/Mermaid.java#L21

ggrossetie:
@tvh Hey! are you still working on it and/or interested to work on it?

Issue #1449, opened 2023-02-21
CORS access blocked because of header


RK-RS:
Hi when trying to get an image (svg or png) from a self hosted kroki instance for a confluence instance sometimes this error appears: 

`Access to XMLHttpRequest at '${KROKI-DIAGRAM-LINK}' from origin '${CONFLUENCE-LINK}' has been blocked by CORS policy: Request header field x-atlassian-mau-ignore is not allowed by Access-Control-Allow-Headers in preflight response.`

is there a way to manually allow this header for CORS through for example setting an env variable in the Docker file?

Running the latest kroki version btw.

ggrossetie:
Indeed, we do not allow the `x-atlassian-mau-ignore` header:

https://github.com/yuzutech/kroki/blob/39017985ebcd0ce9e72ff34140bdc79816fd270b/server/src/main/java/io/kroki/server/Server.java#L93-L97

Currently, this setting is static but I would gladly merge a pull request to make it configurable.

RK-RS:
Added pull request: https://github.com/yuzutech/kroki/pull/1468

PR #1468, opened 2023-03-14
Add ability to add allowed headers to CORS by Environment Variable

RK-RS:
Pull request for this Issue: https://github.com/yuzutech/kroki/issues/1449#issuecomment-1440616969

### DIFF
--- a/server/src/main/java/io/kroki/server/Server.java
+++ b/server/src/main/java/io/kroki/server/Server.java
@@ -48,6 +48,7 @@ import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.handler.BodyHandler;
 import io.vertx.ext.web.handler.CorsHandler;
 
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/kroki/server/Server.java
```diff
@@ -90,11 +92,21 @@ static void start(Vertx vertx, JsonObject config, Handler<AsyncResult<HttpServer
...
 
     // CORS
+    // CORS Headers
     Set<String> allowedHeaders = new HashSet<>();
     allowedHeaders.add("Access-Control-Allow-Origin");
     allowedHeaders.add("Origin");
     allowedHeaders.add("Content-Type");
     allowedHeaders.add("Accept");
+    // Set additional Headers provided by environment variable
+    String envHeadersVar = System.getenv("ALLOWED_HEADERS");
```
ggrossetie, line 102: Please use the `config` object.
I think the option name should be `KROKI_CORS_ALLOWED_HEADERS`.

ggrossetie: We are using [Vert.x Config](https://vertx.io/docs/vertx-config/java/) which can be configured using environment variables or a configuration file.

So, if I set `KROKI_BODY_LIMIT=1234`, it will be available in the `config` object. Take a look at line 92:


```java
config.getLong("KROKI_BODY_LIMIT", BodyHandler.DEFAULT_BODY_LIMIT)
```

So it should be:

```suggestion
    String envHeadersVar = config.getString("KROKI_CORS_ALLOWED_HEADERS");
```

RK-RS: Ahh ok understood, I changed it accordingly. Thanks.

copyrights:
`KROKI_CORS_ALLOWED_HEADERS` should then be add to the documentation (related to #1482). I suggest as part of this pull request.

ggrossetie:
@RK-RS Thanks! Are you comfortable adding a test case for this feature? You can take a look at https://github.com/yuzutech/kroki/blob/main/server/src/test/java/io/kroki/server/ServerUriResponseTest.java which starts the server/deploys the verticle and make an HTTP query using `WebClient`.

Since you can set `DeploymentOptions`, you can start the server using the default configuration and check the response header and do another test where you start the server using something like:

```java
 DeploymentOptions options = new DeploymentOptions()
  .setConfig(new JsonObject()
    .put("KROKI_PORT", port)
    .put("KROKI_CORS_ALLOWED_HEADERS", "x-atlassian-mau-ignore,x-version")
  );
```

ggrossetie:
Merged as part of https://github.com/yuzutech/kroki/commit/47f5b7289f32ba18c0c61a844dd6363a52e1735f (in https://github.com/yuzutech/kroki/pull/1489)

PR #1469, opened 2023-03-14
Documentation: clarify use of boolean diagram options

### LINKED ISSUE #1465: Diagram option 'no-doctype': strange behaviour
As documented [here](https://docs.kroki.io/kroki/setup/diagram-options/#_blockdiag), you can suppress the doctype definition tag for your `blockdiag` diagram (`svg`) using the `no-doctype` diagram option. This works for me only partly, however:

```
$ http https://kroki.io/ diagram_type='actdiag' output_format='svg' diagram_source='{lane Sysadmin}' diagram_options:='{"no-doctype": true  }'
```

This correctly outputs an `svg` image **without** doctype definition tag.

Now I set option `no-doctype` to `false`:

```
$ http https://kroki.io/ diagram_type='actdiag' output_format='svg' diagram_source='{lane Sysadmin}' diagram_options:='{"no-doctype": false  }'
```
**Expected behaviour**: with `no-doctype` set to `false`, doc-type should be shown.
**Actual behaviour**: doc-type is **not** shown.

Is this the intended behaviour?

deining:
This PR addresses and closes #1465 by clarifying the use of boolean diagram options `antialias`, `no-doctype` and `no-transparency`.


docs/modules/setup/pages/diagram-options.adoc
```diff
@@ -12,13 +12,15 @@ Applies to `blockdiag`, `seqdiag`, `actdiag`, `nwdiag`, `packetdiag` and `rackdi
 
 |antialias
 |_boolean_ +
-`true`, *`false`*
-|Pass diagram image to anti-alias filter
+`true`, `false`
+|Apply anti-alias filter to diagram image +
+Option takes effect with both `true` and `false`
```
ggrossetie, line 17: Maybe we should say that the type is `_flag_` and that the option takes effect if the flag is present/specify/set. The value is actually ignored.

What do you think?

deining: > Maybe we should say that the type is `_flag_` and that the option takes effect if the flag is present/specify/set. The value is actually ignored.
> 
> What do you think?

Great idea. I reworked the PR, looks more natural to me now.

docs/modules/setup/pages/diagram-options.adoc
```diff
@@ -11,24 +11,27 @@ Applies to `blockdiag`, `seqdiag`, `actdiag`, `nwdiag`, `packetdiag` and `rackdi
 |Description
 
 |antialias
-|_boolean_ +
-`true`, *`false`*
-|Pass diagram image to anti-alias filter
+|_flag_ +
+empyt string ("")
```
ggrossetie, line 15:
```suggestion
empty string ("")
```

deining: Thanks for spotting this. Fixed.

ggrossetie (approved):
Perfect, thanks👍🏻

PR #1473, opened 2023-03-15
Fix: Kroki server, structurizr: remove output format JPEG

deining:
When sending a request for a `structurizr` diagram with output format `jpeg`, I'm presented a `Error 500`.
I don't think that `structurizr` supports `jpeg` output format, so we should remove it.

### DIFF
--- a/ci/tasks/update-versions.js
+++ b/ci/tasks/update-versions.js
@@ -205,6 +205,15 @@ try {
     }
   }
 
+  const d2GoModContent = await fs.readFile(ospath.join(rootDir, 'server', 'ops', 'docker', 'go.mod'), 'utf8')
+  for (const line of d2GoModContent.split('\n')) {
+    const d2VersionFound = line.match(/^require oss.terrastruct.com\/d2 v(?<version>.+)$/)
+    if (d2VersionFound) {
+      const { version } = d2VersionFound.groups
+      diagramLibraryVersions.d2 = version
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


ggrossetie:
Indeed, I get the following error:

```
java.lang.IllegalArgumentException: No enum constant net.sourceforge.plantuml.FileFormat.JPEG
	at java.base/java.lang.Enum.valueOf(Unknown Source)
	at net.sourceforge.plantuml.FileFormat.valueOf(FileFormat.java:64)
	at io.kroki.server.format.FileFormat.toPlantumlFileFormat(FileFormat.java:16)
	at io.kroki.server.service.Plantuml.convert(Plantuml.java:282)
	at io.kroki.server.service.Structurizr.convert(Structurizr.java:143)
```

I don't know why but I assumed that PlantUML has support for JPEG but apparently that's not the case: https://github.com/plantuml/plantuml/blob/aaf3a8ee247b569bc6d600bd828e80ca35f4db89/src/net/sourceforge/plantuml/FileFormat.java

So we should remove JPEG from PlantUML, Structurizr and C4 PlantUML and update https://kroki.io/ accordingly.


On a side note, PlantUML has a PDF output! If it's working fine, I think we should add it.

ggrossetie:
Related: https://github.com/yuzutech/kroki.io/pull/115

ggrossetie:
It seems that native images of PlantUML do not support PDF output. I will fix it in: https://github.com/ggrossetie/plantuml/blob/de5a47faace1f019c658b75cab019037a86b2c99/.github/workflows/native-image-on-demand.yml#L95-L101

ggrossetie:
Fixed in https://github.com/yuzutech/kroki/pull/1536

PR #1487, opened 2023-03-25
Build arm64 main image

FelixVanOost:
Builds an `arm64` variant of the main Kroki Docker image by leveraging the changes introduced in #1476. All images should now natively support the ARM architecture. This includes the following changes:

- Creates a `nightly` workflow to build the multi-arch images every night at 02:00 UTC.
- Restricts the Docker Buildx builder to 2 parallel threads. This matches the 2 cores available on the GitHub CI runners and prevents timeouts during the build process.
- Runs a [free disk space](https://github.com/jlumbroso/free-disk-space) action at the start of the CI workflow to recover extra disk space from the CI runner. This is now necessary to successfully complete the build process.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,11 +1,5 @@
 name: CI
 
-inputs:
-  build_multiarch:
-    default: 'false'
-    required: true
-    type: boolean
-
 on:
   push:
     paths-ignore:
@@ -18,6 +12,11 @@ on:
     branches:
       - '*'
   workflow_call:
+    inputs:
+      build_multiarch:
+        default: 'false'
+        required: true
+        type: boolean
 
 jobs:
   test-containers:
@@ -26,7 +25,7 @@ jobs:
     - uses: actions/checkout@v3
     
     - name: Free up disk space
-      if: ${{ github.event.inputs.build_multiarch == 'true' }}
+      if: ${{ inputs.build_multiarch }}
       uses: jlumbroso/free-disk-space@main
       with:
         android: true
@@ -48,15 +47,15 @@ jobs:
       run: make buildServer
 
     - name: Set up QEMU
-      if: ${{ github.event.inputs.build_multiarch == 'true' }}
+      if: ${{ inputs.build_multiarch }}
       uses: docker/setup-qemu-action@v2
 
     - name: Set up Docker Buildx
-      if: ${{ github.event.inputs.build_multiarch == 'false' }}
+      if: ${{ !inputs.build_multiarch }}
       uses: docker/setup-buildx-action@v2
 
     - name: Set up Docker Buildx
-      if: ${{ github.event.inputs.build_multiarch == 'true' }}
+      if: ${{ inputs.build_multiarch }}
       uses: docker/setup-buildx-action@v2
       with:
         config-inline: |
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -24,6 +24,7 @@ jobs:
     - uses: actions/checkout@v3
     
     - name: Free up disk space
+      if: ${{ github.event.inputs.build_multiarch == 'true' }}
       uses: jlumbroso/free-disk-space@main
       with:
         android: true
@@ -45,9 +46,15 @@ jobs:
       run: make buildServer
 
     - name: Set up QEMU
+      if: ${{ github.event.inputs.build_multiarch == 'true' }}
       uses: docker/setup-qemu-action@v2
 
     - name: Set up Docker Buildx
+      if: ${{ github.event.inputs.build_multiarch == 'false' }}
+      uses: docker/setup-buildx-action@v2
+
+    - name: Set up Docker Buildx
+      if: ${{ github.event.inputs.build_multiarch == 'true' }}
       uses: docker/setup-buildx-action@v2
       with:
         config-inline: |
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -14,6 +14,7 @@ on:
   workflow_call:
     inputs:
       build_multiarch:
+        default: 'false'
         required: true
         type: boolean
 
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,5 +1,11 @@
 name: CI
 
+inputs:
+  build_multiarch:
+    default: 'false'
+    required: true
+    type: boolean
+
 on:
   push:
     paths-ignore:
@@ -12,11 +18,6 @@ on:
     branches:
       - '*'
   workflow_call:
-    inputs:
-      build_multiarch:
-        default: 'false'
-        required: true
-        type: boolean
 
 jobs:
   test-containers:
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -40,6 +40,10 @@ jobs:
 
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@v2
+      with:
+        config-inline: |
+          [worker.oci]
+            max-parallelism = 2
 
     - name: Cache Docker layers
       uses: actions/cache@v3
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -11,6 +11,11 @@ on:
       - 'docs/**'
     branches:
       - '*'
+  workflow_call:
+    inputs:
+      build_multiarch:
+        required: true
+        type: boolean
 
 jobs:
   test-containers:
@@ -47,6 +52,7 @@ jobs:
     - name: Build container images
       run: make buildDockerImages
       env:
+        BUILD_MULTIARCH: ${{ inputs.build_multiarch }}
         CACHE_FROM: 'type=local,src=/tmp/.buildx-cache'
         CACHE_TO: 'type=local,dest=/tmp/.buildx-cache-new'
 
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -22,6 +22,15 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v3
+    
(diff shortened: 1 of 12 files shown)


ggrossetie:
Could you please rebase on main? Thanks!

FelixVanOost:
The images are building locally but I've been running into several issues with the CI:

- The pipeline was initially reporting a Docker error 317 due to an issue with `cargo install` needing too long / too much memory to get the crates.io index as described in [this](https://github.com/rust-lang/cargo/issues/10781#issuecomment-1163819998) issue. I resolved this by upgrading to the latest Rust release and enabling [sparse-registry](https://blog.rust-lang.org/2022/06/22/sparse-registry-testing.html).
- Next we starting getting some strange network issues with `npm install`, which I resolved by setting `maxsockets=1` but is probably not doing the pipeline any favours in terms of speed.
- Now, it's failing because of a timeout with many of the tests. No idea what's causing this but I will continue looking later.

The full pipeline now takes over an hour because it has to build both `arm64` (using QEMU, which is slower than native) and `amd64` builds. Not ideal, and will probably need to be resolved through more static images.

ggrossetie:
> The full pipeline now takes over an hour because it has to build both arm64 (using QEMU, which is slower than native) and amd64 builds. Not ideal, and will probably need to be resolved through more static images.

Indeed 🙀 
However, I don't think we need to build both arm64 and amd64 on every commit. We could setup a nightly build (with both amd64 and arm64) and a CI build with only amd64.

> The images are building locally but I've been running into several issues with the CI:

We were already close to the GitHub runner limit and I guess adding the arm64 build tipped the balance.
Anyway, thanks for finding workarounds and improvements 👍🏻 

Also it might be possible to reduce the parallelization of `buildx` by running it multiple times on different targets:


```
docker buildx bake kroki-python
docker buildx bake kroki-js
docker buildx bake kroki-java
```

FelixVanOost:
I've defined a nightly CI workflow that will build the multi-arch images and also split up the `docker buildx bake` calls for the main vs. companion images to reduce parallelisation. All of this works fine locally, but we're still getting a timeout error for some of the tests defined in `/vega/tests/test.js`.

@ggrossetie I couldn't figure out where the 2000ms timeout is defined for these tests. Is there a global config somewhere? I think that the problem would be resolved by increasing the timeout slightly, because running the `arm64` image on QEMU is slower than native.

ggrossetie:
> I couldn't figure out where the 2000ms timeout is defined for these tests. Is there a global config somewhere? I think that the problem would be resolved by increasing the timeout slightly, because running the arm64 image on QEMU is slower than native.

I believe this is the default value of Mocha but you can override it using `this.timeout(15000)` (15 seconds in milliseconds) inside the `describe` function in https://github.com/yuzutech/kroki/blob/b039f177440f57bc93c0ff07d5feb633b7a9b7c3/vega/tests/test.js#L15

FelixVanOost:
After resolving the parallelism issues (I think), I've now encountered issues with running out of disk space on the runners. For now I'm trying to run this [action](https://github.com/jlumbroso/free-disk-space) at the start of the workflow, and I'll see if that fixes things.

@ggrossetie Please let me know if you'd like me to stop this; I don't want to waste all your CI minutes. 🙃

FelixVanOost:
This is now good to go from my side. I've updated the PR description above with the main changes.

I tested both regular (native-arch) and multi-arch CI workflows manually in the two last commits. Nightly and release builds are multi-arch, regular builds are native-arch.

ggrossetie:
@FelixVanOost perfect, I will take a closer look this weekend 👍🏻

ggrossetie:
I will try to build the arm64 image on an [Ampere A1 Compute](https://www.oracle.com/fr/cloud/compute/arm/) machine. I heard that QEMU is ~20 times slower.

ggrossetie:
Also, in the next version we will use UMLet and PlantUML as native images (using GraalVM). So we need to produce arm64 compatible native images.

.github/workflows/main.yaml
```diff
@@ -11,12 +11,26 @@ on:
...
     - uses: actions/checkout@v3
+    
+    - name: Free up disk space
+      uses: jlumbroso/free-disk-space@main
+      with:
+        android: true
+        dotnet: true
+        haskell: true
+        large-packages: true
+        swap-storage: true
```
ggrossetie, line 33: I think we should use a condition to use this action only if `build_multiarch` is `true`

.github/workflows/main.yaml
```diff
@@ -35,6 +49,10 @@ jobs:
 
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@v2
+      with:
+        config-inline: |
+          [worker.oci]
+            max-parallelism = 2
```
ggrossetie, line 55: Same here,we should set max-parallelism to 2 when `build_multiarch` is `true`

ggrossetie:
Thanks for all your great work @FelixVanOost 🙌🏻

FelixVanOost:
My pleasure!

FelixVanOost:
> I will try to build the arm64 image on an [Ampere A1 Compute](https://www.oracle.com/fr/cloud/compute/arm/) machine. I heard that QEMU is ~20 times slower.

A native `arm64` runner would be great. I ended up using cross-compilation (see [here](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/)) for most of the builds, so the CI build times should be much lower than they were initally.

QEMU is now only being used for D2, Pikchr, Mermaid, Wireviz, and Blockdiag. Of those, D2, Pikchr, and Mermaid can most likely be cross-compiled with some additional effort.

ggrossetie:
@FelixVanOost I get the following errors when building images:

```
WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
ERROR: (*service).Write failed
```

Not sure if it means that the disk is full or if the latest version of buildx has a bug...? Does this error sound familiar?

ggrossetie:
Apparently, when something is failing it throws this error but this is not the root cause

FelixVanOost:
> @FelixVanOost I get the following errors when building images:
> 
> ```
> WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
> WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
> WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
> WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
> WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
> WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
> WARNING: local cache import at /tmp/.buildx-cache not found due to err: could not read /tmp/.buildx-cache/index.json: open /tmp/.buildx-cache/index.json: no such file or directory
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ERROR: (*service).Write failed
> ```
> 
> Not sure if it means that the disk is full or if the latest version of buildx has a bug...? Does this error sound familiar?

I've never experienced this before. Is this happening directly in the workflow?

mariotoffia:
@FelixVanOost  thanks for you're porting kroki (of which I'm ever grateful @ggrossetie) to arm64!!

Cheers,
 Mario :)

PR #1511, opened 2023-04-14
Documentation: complete available themes for d2 diagrams

deining:
This PR adds further themes available for d2 diagrams to the documentation.
Were great if the docs could be redeployed after merge so that changes take effect.

### DIFF
--- a/docs/modules/setup/pages/diagram-options.adoc
+++ b/docs/modules/setup/pages/diagram-options.adoc
@@ -58,8 +58,11 @@ With flag set, option takes effect.
 * `earth-tones` (`103`)
 * `everglade-green` (`104`)
 * `buttered-toast` (`105`)
+* `dark-mauve` (`200`)
+* `terminal` (`300`)
+* `terminal-grayscale` (`301`)
 
-| See: https://github.com/terrastruct/d2/blob/master/docs/assets/themes_overview.png
+| See: https://d2lang.com/tour/themes/
 |===
 
 == GraphViz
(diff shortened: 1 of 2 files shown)


ggrossetie:
Before that we need to update D2 and support the additional themes: 
https://github.com/yuzutech/kroki/blob/ca6442c21c78888334bc6be5931a6149115dedf6/server/src/main/java/io/kroki/server/service/D2.java#L30-L45

deining:
> Before that we need to update D2 and support the additional themes:

To me it look like these themes are supported already: send a request for an arbitrary d2 svg digram. As diagram option, specify `theme = 200`. This will return back an image with `Dark mauve` theme applied correctly.

Also: iIf you give an invalid numeric value for the theme (e.g. theme = `42`), the error message is:

Internal Server Error. Reason: Error 500: err: bad usage: -t[heme] could not be found. The available options are: Light: - Neutral default: 0 err: - Neutral Grey: 1 err: - Flagship Terrastruct: 3 err: - Mixed berry blue: 5 err: - Cool classics: 4 err: - Grape soda: 6 err: - Aubergine: 7 err: - Colorblind clear: 8 err: - Vanilla nitro cola: 100 err: - Orange creamsicle: 101 err: - Shirley temple: 102 err: - Earth tones: 103 err: - Everglade green: 104 err: - Buttered toast: 105 err: - Terminal: 300 err: - Terminal Grayscale: 301 err: Dark: - Dark Mauve: 200 err: You provided: 42 err: Run with --help to see usage.  (exit code 1).

This also indicates, that the current d2 version on the server already knows about/supports the additional themes that I would like to see documented.

ggrossetie:
> This also indicates, that the current d2 version on the server already knows about/supports the additional themes that I would like to see documented.

I didn't know that we are already supporting "dark mauve", however we still need to update the code to add the mapping between the numeric value and the name.
If you use `?theme=dark-mauve` it won't work.

ggrossetie:
Perfect, thanks 🙌🏻

PR #1515, opened 2023-04-16
resolves #1480 Use PlantUML as native image

### LINKED ISSUE #1480: Call PlantUML as a native image rather than a Java library
The goal is to improve the stability of the gateway server and provide better isolation. Each PlantUML request will be executed by a dedicated process.

We can use agressive timeout/limit on each process to make sure that a specific request won't consume too much ressources.

See: https://github.com/plantuml/plantuml/issues/1344

ggrossetie:
- closes #1480 

#### TODO

- [x] Implement PNG/base64 (maybe we should add it on all diagram librairies which support PNG)
- [x] Enable Structurizr and C4 PlantUML tests again

#### Known issues

- `help themes` won't list all the themes


ggrossetie:
> Implement PNG/base64 (maybe we should add it on all diagram librairies which support PNG)

I'm actually not quite sure about this feature. We might want to implement it as a post-processor step since technically we could produce base64 output from PNG, or JPEG or even SVG (and/or PDF?).
In my opinion, it does not really make sense for utxt or txt output but who knows... 

Anyway, I will restore this feature as-is for now.

Issue #1512, opened 2023-04-15
Pass SSL environment variables as paths instead of strings


FelixVanOost:
The environment variables `KROKI_SSL_KEY` and `KROKI_SSL_CERT` currently store the contents of the key and certificate as strings, which differs from the more common usage in which the file paths are provided directly. The current approach has two drawbacks that I can see:

- The environment variables must be manually updated whenever the TLS certificate is renewed. This is typically every 3 months (e.g., for Let's Encrypt and ZeroSSL).
- The certificate and private key are accessible by any system user via the environment variables or, if defined in a `docker-compose.yml`, anyone with access to the file.

Both of these issues could be resolved by storing paths to the certificate and key in the environment variables instead and mounting a volume containing the files to the main Docker image.

ggrossetie:
I can support this idea if we keep the existing feature as both options have advantages and drawbacks.
For instance, if you declare key and certificats as paths, you will need to make sure that the container has access to these files (i.e., mount a local disk or copy files inside the container).

In my opinion, we should add two new environment variables:

- `KROKI_SSL_KEY_PATH`
- `KROKI_SSL_CERT_PATH`

I think `KROKI_SSL_KEY` and `KROKI_SSL_CERT` should have higher precedence. Also make sure to update the documentation to document this feature.

PR #1517, opened 2023-04-19
Allow SSL certificate to be passed as a file path

FelixVanOost:
Allows the SSL certificate and private key to be passed to the Docker image as a file path instead of a string (see #1512). This includes the following changes:

- Adds two new environment variables `KROKI_SSL_KEY_PATH` and `KROKI_SSL_CERT_PATH`. These are given lower priority than the existing `KROKI_SSL_KEY` and `KROKI_SSL_CERT` variables in case of multiple definitions.
- Adds new server tests to check for correct SSL setup in both possible cases (certificate + key provided as strings or as paths).
- Updates the documentation to describe both possible options.

### DIFF
--- a/docs/modules/setup/pages/configuration.adoc
+++ b/docs/modules/setup/pages/configuration.adoc
@@ -271,4 +271,4 @@ podman run -p8000:8000 -e KROKI_SSL=true --env-file=.env yuzutech/kroki
 
 ====
 
-If SSL is enabled, both `KROKI_SSL_KEY` / `KROKI_SSL_KEY_PATH` and `KROKI_SSL_CERT` / `KROKI_SSL_CERT_PATH` must be configured.
+If SSL is enabled, both `KROKI_SSL_KEY` (or `KROKI_SSL_KEY_PATH`) and `KROKI_SSL_CERT` (or `KROKI_SSL_CERT_PATH`) must be configured.
--- a/docs/modules/setup/pages/configuration.adoc
+++ b/docs/modules/setup/pages/configuration.adoc
@@ -215,7 +215,23 @@ We recommend to use a maximum length that's not greater than 8192 and not greate
 
 By default, SSL/TLS is not enabled on the server but you can enable it by setting `KROKI_SSL` environment variable to `true`.
 
-When SSL is enabled, you must provide the certificate and the private key values as PEM format using `KROKI_SSL_KEY` and `KROKI_SSL_CERT` environment variables.
+When SSL is enabled, you must provide the certificate and the private key in one of two ways:
+
+- As strings in PEM format using the `KROKI_SSL_KEY` and `KROKI_SSL_CERT` environment variables, e.g.,:
++
+[source,bash]
+----
+KROKI_SSL_KEY="-----BEGIN RSA PRIVATE KEY-----<PRIVATE_KEY>-----END RSA PRIVATE KEY-----"
+----
+
+- As PEM file paths using the `KROKI_SSL_KEY_PATH` and `KROKI_SSL_CERT_PATH` environment variables.
++
+[source,bash]
+----
+KROKI_SSL_KEY_PATH="/etc/ssl/certs/mydomain/privatekey.pem"
+----
+
+If both methods are used, the values in `KROKI_SSL_KEY` and `KROKI_SSL_CERT` are given priority.
 
 [NOTE]
 ====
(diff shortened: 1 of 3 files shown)


docs/modules/setup/pages/configuration.adoc
```diff
@@ -255,5 +271,4 @@ podman run -p8000:8000 -e KROKI_SSL=true --env-file=.env yuzutech/kroki
 
 ====
 
-If SSL is enabled, both `KROKI_SSL_KEY` and `KROKI_SSL_CERT` must be configured.
-
+If SSL is enabled, both `KROKI_SSL_KEY` / `KROKI_SSL_KEY_PATH` and `KROKI_SSL_CERT` / `KROKI_SSL_CERT_PATH` must be configured.
```
ggrossetie, line 274: I would say:

```suggestion
If SSL is enabled, both `KROKI_SSL_KEY` and `KROKI_SSL_CERT` (or `KROKI_SSL_KEY_PATH` and `KROKI_SSL_CERT_PATH`) must be configured.
```

ggrossetie: Oh I see, we could use `KROKI_SSL_KEY_PATH` with `KROKI_SSL_CERT`?

ggrossetie: then it should be:

```suggestion
If SSL is enabled, both `KROKI_SSL_KEY` (or `KROKI_SSL_KEY_PATH`) and `KROKI_SSL_CERT` (or `KROKI_SSL_CERT_PATH`) must be configured.
```

server/src/main/java/io/kroki/server/Server.java
```diff
@@ -178,16 +178,29 @@ static void start(Vertx vertx, JsonObject config, Handler<AsyncResult<HttpServer
...
+
+      if (!config.containsKey("KROKI_SSL_KEY") && !config.containsKey("KROKI_SSL_KEY_PATH")) {
+        throw new IllegalArgumentException("KROKI_SSL_KEY or KROKI_SSL_KEY_PATH must be configured when SSL is enabled.");
       }
-      serverOptions.setPemKeyCertOptions(
-        new PemKeyCertOptions()
-          .addKeyValue(Buffer.buffer(sslKeyValue.get()))
-          .addCertValue(Buffer.buffer(sslCertValue.get()))
-      );
+      else if (config.containsKey("KROKI_SSL_KEY") && !config.getString("KROKI_SSL_KEY").isEmpty()) {
```
ggrossetie, line 186: we don't need `else if` when the `if` block returns a value or throws an error

```suggestion
      if (config.containsKey("KROKI_SSL_KEY") && !config.getString("KROKI_SSL_KEY").isEmpty()) {
```

server/src/main/java/io/kroki/server/Server.java
```diff
@@ -178,16 +178,29 @@ static void start(Vertx vertx, JsonObject config, Handler<AsyncResult<HttpServer
...
+        certOptions.addKeyValue(Buffer.buffer(config.getString("KROKI_SSL_KEY")));
+      }
+      else {
+        certOptions.addKeyPath(config.getString("KROKI_SSL_KEY_PATH"));
+      }
+
+      if (!config.containsKey("KROKI_SSL_CERT") && !config.containsKey("KROKI_SSL_CERT_PATH")) {
+        throw new IllegalArgumentException("KROKI_SSL_CERT or KROKI_SSL_CERT_PATH must be configured when SSL is enabled.");
+      }
```
ggrossetie, line 195: I would prefer to move this block at the start of the block. I like to have all "preconditions" at the top.

server/src/main/java/io/kroki/server/Server.java
```diff
@@ -178,16 +178,29 @@ static void start(Vertx vertx, JsonObject config, Handler<AsyncResult<HttpServer
 
   private static void setPemKeyCertOptions(JsonObject config, HttpServerOptions serverOptions, boolean enableSSL) {
     if (enableSSL) {
-      Optional<String> sslKeyValue = Optional.ofNullable(config.getString("KROKI_SSL_KEY"));
-      Optional<String> sslCertValue = Optional.ofNullable(config.getString("KROKI_SSL_CERT"));
-      if (sslKeyValue.isEmpty() || sslCertValue.isEmpty()) {
-        throw new IllegalArgumentException("KROKI_SSL_KEY and KROKI_SSL_CERT must be configured when SSL is enabled.");
+      PemKeyCertOptions certOptions = new PemKeyCertOptions();
+
+      if (!config.containsKey("KROKI_SSL_KEY") && !config.containsKey("KROKI_SSL_KEY_PATH")) {
```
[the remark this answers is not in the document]
FelixVanOost: I've made and pushed the changes.

ggrossetie:
Looks good, thanks 👍🏻

PR #1522, opened 2023-04-23
Use pre-built binary for GraphViz

### LINKED ISSUE #1483: Build Graphviz from source
The latest version of Graphviz available via `apt` ([2.42.2](https://packages.ubuntu.com/jammy/graphviz)) was first released in September 2019 and is very out of date. From the Graphviz [Download](https://graphviz.org/download/) page, the maintainers note that they no longer provide precompiled packages.

By building Graphviz from source, it should be easy to update to much newer releases (which are made every month or so).

ggrossetie:
- closes #1483

### DIFF
--- a/ci/tasks/update-versions.js
+++ b/ci/tasks/update-versions.js
@@ -198,6 +198,11 @@ try {
       const { version } = plantumlVersionFound.groups
       diagramLibraryVersions.plantuml = version
     }
+    const graphvizVersionFound = line.match(/^ARG GRAPHVIZ_VERSION="(?<version>.+)"$/)
+    if (graphvizVersionFound) {
+      const { version } = graphvizVersionFound.groups
+      diagramLibraryVersions.graphviz = version
+    }
   }
 
   const svgbobCargoContent = await fs.readFile(ospath.join(rootDir, 'server', 'ops', 'docker', 'Cargo.toml'), 'utf8')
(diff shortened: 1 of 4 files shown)


ggrossetie:
The `deb` file is only for `amd64`:

```
dpkg: error processing archive graphviz.deb (--install):
 package architecture (amd64) does not match system (arm64)
Errors were encountered while processing:
 graphviz.deb
```

FelixVanOost:
> The `deb` file is only for `amd64`:
> 
> ```
> dpkg: error processing archive graphviz.deb (--install):
>  package architecture (amd64) does not match system (arm64)
> Errors were encountered while processing:
>  graphviz.deb
> ```

We could try to get this resolved upstream, but they already release packages for a large number of distributions so might be reluctant to double the number of CI jobs they run...

ggrossetie:
> We could try to get this resolved upstream, but they already release packages for a large number of distributions so might be reluctant to double the number of CI jobs they run...

Yes, we can open an issue but as you said they might be unwilling to do so. I now have a dedicated GitHub runner that runs on ARM64. I will try to setup a job to produce the `graphviz.deb` for ARM64.

As a side note, smoke tests indicate that something does not work as expected with GraphViz 8.0.4. Not sure if they follow semantic versioning but that's quite a big jump from 2.42 (or 3.0) to 8.0 😅 
Since this is the latest version, I'm willing to accommodate the code to make it work with GraphViz 8+

ggrossetie:
I need to build GraphViz on arm64 and then we are good to go :)

FelixVanOost:
Nice! 😎

PR #1523, opened 2023-04-24
Install arm64-compatible Node packages

FelixVanOost:
Fixes an issue in #1487 where `amd64` packages were installed for all the Node-based diagram libraries when building for the `arm64` architecture. This prevented any of those diagram libraries from working on native `arm64` machines (without QEMU / Rosetta 2).

I have tested that these changes resolve the issue on a native `arm64` machine and that all the Node-based libraries work as expected.


FelixVanOost:
@ggrossetie This wasn't caught by the CI because the smoke tests aren't currently executed with the `arm64` images. We can look into running a separate workflow for this now that you have a dedicated runner available.

PR #1536, opened 2023-05-07
Replace JPEG by PDF for PlantUML, C4 PlantUML and Structurizr

ggrossetie:


server/ops/docker/jdk11-jammy/Dockerfile
```diff
@@ -211,6 +211,7 @@ ARG D2_VERSION="0.4.1"
 ARG PLANTUML_VERSION="1.2023.6"
 ARG UMLET_VERSION="2023-03-20_UMLet_v15.1"
 ARG GRAPHVIZ_VERSION="8.0.4"
+ARG GRAPHVIZ_VERSION="8.0.4"
```
ggrossetie, line 214:
```suggestion
```

ggrossetie:
Thanks @deining for initiating this change.
I had a few issues with PDF output with the native image of PlantUML but it seems to be good now 😌

PR #1540, opened 2023-05-08
Remove deprecated installLocalDependencies workflow step

FelixVanOost:
Removes the `installLocalDependencies` step from the `release` workflow and README after being deprecated in #1537.


ggrossetie:
Ooops, good catch 😅 
Thanks for having my back!

PR #1548, opened 2023-05-19
Add missing Docker arguments for arm64 builds

FelixVanOost:
Adds missing `TARGETARCH` and `TARGETOS` arguments to the BPMN, Diagrams.net, and Excalidraw Dockerfiles to fix the broken `arm64` builds of these images. Without these arguments, the binaries are always built for the host architecture (currently `amd64`) instead of the target.

### DIFF
--- a/bpmn/Dockerfile
+++ b/bpmn/Dockerfile
@@ -8,9 +8,10 @@ RUN npm config set unsafe-perm true
 
 RUN npm install -g pkg@5.6.0 pkg-fetch@3.3.0
 
+ARG TARGETARCH
+ARG TARGETOS
 ENV NODE node16
-ENV PLATFORM alpine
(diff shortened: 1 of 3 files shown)


bpmn/Dockerfile
```diff
@@ -19,7 +20,7 @@ WORKDIR /app
 RUN PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 npm ci
 RUN npm run lint
 
-RUN /usr/local/bin/pkg --targets ${NODE}-${PLATFORM}-$([ "$TARGETARCH" == "amd64" ] && echo "x64" || echo "$TARGETARCH") . -o app.bin
```
ggrossetie, line 22: We need to keep `alpine` as "platform". `pkg` makes a distinction between linux and alpine:

> **platform** alpine, linux, linuxstatic, win, macos, (freebsd)

https://github.com/vercel/pkg#targets

ggrossetie:
Thanks!

PR #1560, opened 2023-06-03
Correct spelling of D2 theme

deining:
The D2 [themes list](https://d2lang.com/tour/themes/), mentions `neutral-gray`, not `neutral-grey`.
I propose we should follow the original nomenclature here.


ggrossetie:
Good catch 👍🏻

Issue #1423, opened 2023-01-08
Service discovery


Mogztter:
The goal is to allow the gateway server to register additional diagram libraries (that are not built-in). 
One of the requirement will be to adhere to the Kroki companion container API specification (not yet available).

#### Discovery

When the gateway server starts, it will probe the seed addresses by connecting to each address and attempting to identify the service. If successful, it will register a new service.

```yaml
discovery.seed_resolver.timeout: 5s
discovery.seed_resolver.max_retry: 5
discovery.seed_hosts:
  - 192.168.1.10:1234
  - 192.168.1.11 
  - seeds.mydomain.com 
```

This solution implies that the service must provide information (at least a unique name, a version, and a list of supported formats).

#### Static registration

Alternatively, we could explicitly register services:

```yaml
services:
  - name: sankey
    version: '1.2.3'
    endpoint: https://my.domain.org/sankey
    formats:
      - png
      - svg
```

In this case, the service does not have to expose an API that provides information.

#### Dynamic registration

We could also support dynamic registration. In this case, the companion service will be responsible for registering itself by sending a request to the gateway server.

```http
POST https://my.kroki/services

{
  "name": "mscgen",
  "version": "1.2.3",
  "formats": ["png", "svg"]
}
```

- `201 created`: service registration is successful
- `422 unprocessable entity`: non-unique name (service is probably already registered)
- `400 bad request`: request payload is invalid
- `500`: unexpected error

Since Kroki is stateless, registry is held in memory. As a result, the companion service is responsible for registering again in case the gateway server is restarted.

For this reasons, services should send a heartbeat:

```http
PUT https://my.kroki/services/mscgen/heartbeat
```

- `200 ok`: service found
- `404 not found`: service is not registered (use `_links` to indicate how to register)
- `500`: unexpected error

A service that receives a 404 during a heartbeat could then register again.

The gateway service will provide information about a registred service:

```http
GET https://my.kroki/services/mscgen

{
  "name": "mscgen",
  "version": "1.2.3",
  "formats": ["png", "svg"],
  "created_at": "2023-01-01T15:50:05.815479Z",
  "updated_at": "2023-01-01T15:50:05.815479Z",
  "last_activity_at": "2023-01-01T18:55:04.187568Z"
}
```

- `200 ok`: service found
- `404 not found`: service is not registered
- `500`: unexpected error

Note: Last activity date is the last time a service sent a heartbeat or respond to a request from the gateway server.

[...]

PR #1566, opened 2023-06-08
Add symbolator

### LINKED ISSUE #1505: Support Hardware Definition Language
Hi, great project!
Please consider supporting Symbolator for generating hardware description language blocks like vhdl or verilog.

zebreus:
This PR will add support for [symbolator](https://hdl.github.io/symbolator/) component diagrams.

resolves #1505

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -37,12 +37,6 @@ jobs:
     - name: Install GraphViz
       run: sudo apt-get install graphviz -y
 
-    - name: Install Symbolator
-      run: |
-        wget https://github.com/zebreus/symbolator/releases/download/v1.2.2/symbolator.AppImage
-        chmod +x symbolator.AppImage
-        mv symbolator.AppImage /usr/local/bin/symbolator
-
     - name: Set up JDK 11
       uses: actions/setup-java@v3
       with:
--- a/ci/tasks/update-versions.js
+++ b/ci/tasks/update-versions.js
@@ -11,7 +11,6 @@ const require = createRequire(import.meta.url)
 const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
 const rootDir = ospath.join(__dirname, '..', '..')
 const krokiServicePath = ospath.join(rootDir, 'server', 'src', 'main', 'java', 'io', 'kroki', 'server', 'service')
-const githubActionsPath = ospath.join(rootDir, '.github', 'workflows')
 
 const diagramLibraryVersions = {}
 
@@ -31,19 +30,6 @@ async function updateServiceGetVersion(javaServiceFileName, version) {
     if (currentVersion !== version) {
       const updatedJavaContent = javaContent.replace(/(?<= +public String getVersion\(\) {\n\s+return ")(?<version>[0-9.]+)(?=";\n\s+})/, version)
       await fs.writeFile(servicePath, updatedJavaContent, 'utf8')
-    }
-  }
-}
-
-async function updateSymbolatorAppImageVersion(githubActionFileName, version) {
-  const actionPath = ospath.join(githubActionsPath, githubActionFileName)
-  const githubActionContent = await fs.readFile(actionPath, 'utf8')
-  const versionFound = githubActionContent.match(/(?<=symbolator\/releases\/download\/v)(?<version>[0-9.]+)(?=\/symbolator\.AppImage)/)
-  if (versionFound) {
-    const currentVersion = versionFound.groups.version
-    if (currentVersion !== version) {
-      const updatedGithubActionContent = githubActionContent.replace(/(?<=symbolator\/releases\/download\/v)(?<version>[0-9.]+)(?=\/symbolator\.AppImage)/, version)
-      await fs.writeFile(actionPath, updatedGithubActionContent, 'utf8')
     }
   }
 }
@@ -284,7 +270,6 @@ try {
   await updateServiceGetVersion('Wavedrom.java', diagramLibraryVersions.wavedrom)
   await updateServiceGetVersion('Wireviz.java', diagramLibraryVersions.wireviz)
   await updateVegaServiceGetVersion(diagramLibraryVersions.vega, diagramLibraryVersions.vegalite)
-  await updateSymbolatorAppImageVersion("main.yaml", diagramLibraryVersions.symbolator)
 
 } catch (err) {
   console.error(err)


ggrossetie:
Woot 🎉 
@zebreus Is it ready for review?

zebreus:
Now it is ready for review.

My changes are mostly based on the existing code for ditaa.

The [upstream of symbolator](https:github.com/kevinpt/symbolator) is broken with newer python versions and unmaintained for 6 years. There was a [updated fork](https:github.com/hdl/symbolator) but that one also seems abandoned for at least a year. VHDL support is also broken at that fork. I [forked it again](https:github.com/hdl/symbolator), merged all open PRs and added a few small features.

The kroki docker build installs symbolator directly from the GitHub source, I also contacted the original author to add me as a maintainer on pypi so in the future it should be able to be installed from there. I packaged symbolator as [AppImage](https://github.com/zebreus/symbolator/releases/tag/v1.2.1) and [docker container](https://hub.docker.com/r/madmanfred/symbolator) these are probably the easiest ways to get a executable for testing.

I added a test to check that symbolator is kind of working, maybe that should be removed as it makes the server build depend on having a symbolator executable.

ggrossetie:
> The [upstream of symbolator](https://github.com/kevinpt/symbolator) is broken with newer python versions and unmaintained for 6 years. There was a [updated fork](https://github.com/hdl/symbolator) but that one also seems abandoned for at least a year. VHDL support is also broken at that fork. I [forked it again](https://github.com/zebreus/symbolator), merged all open PRs and added a few small features.

That's not reassuring 😨 
I didn't see that... if the project is unmaintained, I don't think we should add it in the "main" distribution. I found a "popular" project named https://github.com/logisim-evolution/logisim-evolution. Maybe they would be interested in working/maintaining a diagram library to render HDL files as images?

I'm planning to work on https://github.com/yuzutech/kroki/issues/1423. That would allow third-party companion containers to register to the gateway server.

server/ops/docker/jdk11-jammy/Dockerfile
```diff
@@ -155,6 +155,34 @@ FROM yuzutech/kroki-builder-erd:0.2.1.0 AS kroki-builder-static-erd
...
+WORKDIR /build
+
+# Install latest pip and setuptools
+RUN python3 -m pip install --upgrade pip setuptools
+RUN python3 -m pip install --upgrade nuitka
+# Install symbolator
+RUN python3 -m pip install --upgrade ${SYMBOLATOR_SOURCE}
+
+# Use nuitka to compile a static binary so we dont need python in the final image
+RUN python3 -m nuitka --onefile `which symbolator` --include-module=gi.overrides.Pango --include-module=gi._gi_cairo
```
ggrossetie, line 184: 👍🏻

.github/workflows/main.yaml
```diff
@@ -37,6 +37,12 @@ jobs:
     - name: Install GraphViz
       run: sudo apt-get install graphviz -y
 
+    - name: Install Symbolator
+      run: |
+        wget https://github.com/zebreus/symbolator/releases/download/v1.2.1/symbolator.AppImage
+        chmod +x symbolator.AppImage
+        mv symbolator.AppImage /usr/local/bin/symbolator
```
ggrossetie, line 44: I don't see any unit tests?

server/ops/docker/jdk11-jammy/Dockerfile
```diff
@@ -155,6 +155,34 @@ FROM yuzutech/kroki-builder-erd:0.2.1.0 AS kroki-builder-static-erd
 #  --ghc-options="-optc=-static" \
 #  --ghc-options="-w"
 
+## Symbolator
+FROM ubuntu:jammy AS kroki-builder-static-symbolator
+
+# Build from forked source because upstream is broken for the latest python3 versions
+ARG SYMBOLATOR_SOURCE=git+https://github.com/zebreus/symbolator.git@v1.2.1
```
ggrossetie, line 162: we should extract the Symbolator version as an `ARG`

docs/antora.yml
```diff
@@ -25,6 +25,7 @@ asciidoc:
     seqdiag-version: 3.0.0
     structurizr-version: 1.30.1
     svgbob-version: 0.7.0
+    symbolator-version: 1.2.1
```
ggrossetie, line 28: Please update the `tasks/update-versions.js` script to automatically update Symbolator version (based on the version defined in the Dockerfile).

We are already doing this for other diagram libraries so you can copy/paste. Let me know if you need help.

server/src/main/java/io/kroki/server/service/SymbolatorCommand.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+    String omitTypeInformation = options.getString("no-type");
+    if (omitTypeInformation != null) {
+      commands.add("--no-type");
+    }
+    String libname = options.getString("libname");
+    if (libname != null) {
+      commands.add("--libname ");
+      commands.add(libname);
+    }
```
ggrossetie, line 77: could you please document these options in https://github.com/yuzutech/kroki/blob/main/docs/modules/setup/pages/diagram-options.adoc

server/ops/docker/jdk11-jammy/Dockerfile
ggrossetie: > Use nuitka to compile a static binary so we dont need python in the final image

@zebreus Do you think it would be possible to do the same with blockdiag (and wireviz)? My initial goal was to create a binary version but I gave up because I didn't find a working solution. If you think it would be possible, I can create an issue to track progress. Thanks!

ggrossetie:
> Maybe I was a bit to negative in my communication about symbolator. The custom version is mostly necessary to get it running with recent versions of python3. Symbolator itself is stable. The other changes are mostly small improvements like cli options to select a component or to get a diagram with a transparent background.

Got it 👍🏻 

> Symbolator was also included in asciidoctor-diagram which was the reason why I used it in the first place. I think it should be in the main distribution of kroki to make transitions to asciidoctor.js and asciidoctor-web-pdf easier.
Symbolator adds about 20MB to the image (from 1795MB to 1815MB).

My main concerns were security and performance (i.e., if we encounter a security or performance issue and nobody can fix it). I thought about providing a way to dynamically disable a particular diagram library.

> I can look into converting this to a companion container when service discovery is ready, but would love to have the changes upstreamed before, as I plan to use it in a project and would rather not host my own kroki server.

👍🏻

ggrossetie:
Looks good, thanks!
I left a few comments that need to be resolved before we can merge.

docs/antora.yml
zebreus: :+1: It now also updates the symbolator version used in github actions based on the Dockerfile.

zebreus:
> > I added a test to check that symbolator is kind of working, maybe that should be removed as it makes the server build depend on having a symbolator executable.
> 
> You can keep the test but it should check that the binary exists.

I put that test along the service tests, but already removed it a few commits ago, as it did not really belong there ([SymbolatorServiceTest.java](https://github.com/zebreus/kroki/blob/45751c4b0bcbd6acd357b6e412fc67b4bb3a2ea2/server/src/test/java/io/kroki/server/service/SymbolatorServiceTest.java)). The smoke tests achieve basically the same thing, except for checking the output.

If you want to I could add a few testcases somewhere else.

server/ops/docker/jdk11-jammy/Dockerfile
zebreus: It seems like nuitka is able to build working static binaries for wireviz without problems. Blockdiag also builds fine but fails at runtime, because it has some missing dependency but that can probably be fixed.

A tracking issue is probably good for further discussion. Do you think we can integrate blockdiag and wireviz into the main container if we have static binaries?

Nuitka also seems to [support multi-call binaries](https://nuitka.net/doc/user-manual.html#use-case-6-multidist) so in the best-case we could build a single binary for all blockdiag packages. Maybe even for all python diagram generators but I am still undecided whether that is a good idea.

ggrossetie:
> If you want to I could add a few testcases somewhere else. 

No I think it's OK.

.github/workflows/main.yaml
```diff
@@ -37,6 +37,12 @@ jobs:
     - name: Install GraphViz
       run: sudo apt-get install graphviz -y
 
+    - name: Install Symbolator
+      run: |
+        wget https://github.com/zebreus/symbolator/releases/download/v1.2.2/symbolator.AppImage
+        chmod +x symbolator.AppImage
+        mv symbolator.AppImage /usr/local/bin/symbolator
+
```
ggrossetie, line 45: We can remove this step right?

```suggestion
```

server/ops/docker/jdk11-jammy/Dockerfile
ggrossetie: > Do you think we can integrate blockdiag and wireviz into the main container if we have static binaries?

That could be an option, yes.

> Nuitka also seems to support multi-call binaries so in the best-case we could build a single binary for all blockdiag packages. Maybe even for all python diagram generators but I am still undecided whether that is a good idea.

One for all blockdiag packages is a good idea!

> A tracking issue is probably good for further discussion.

I will create one or since you've already looked into it feel free to create one.

ci/tasks/update-versions.js
```diff
@@ -34,7 +35,20 @@ async function updateServiceGetVersion (javaServiceFileName, version) {
...
+  const versionFound = githubActionContent.match(/(?<=symbolator\/releases\/download\/v)(?<version>[0-9.]+)(?=\/symbolator\.AppImage)/)
+  if (versionFound) {
+    const currentVersion = versionFound.groups.version
+    if (currentVersion !== version) {
+      const updatedGithubActionContent = githubActionContent.replace(/(?<=symbolator\/releases\/download\/v)(?<version>[0-9.]+)(?=\/symbolator\.AppImage)/, version)
+      await fs.writeFile(actionPath, updatedGithubActionContent, 'utf8')
+    }
+  }
+}
+
```
ggrossetie, line 50:
```suggestion
```

ci/tasks/update-versions.js
```diff
@@ -11,17 +11,18 @@ const require = createRequire(import.meta.url)
 const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
 const rootDir = ospath.join(__dirname, '..', '..')
 const krokiServicePath = ospath.join(rootDir, 'server', 'src', 'main', 'java', 'io', 'kroki', 'server', 'service')
+const githubActionsPath = ospath.join(rootDir, '.github', 'workflows')
```
ggrossetie, line 14:
```suggestion
```

ci/tasks/update-versions.js
```diff
@@ -11,17 +11,18 @@ const require = createRequire(import.meta.url)
...
-function getDependencyVersion (pkg, name) {
+function getDependencyVersion(pkg, name) {
   if (name in pkg.dependencies) {
     return pkg.dependencies[name]
   }
   return pkg.devDependencies[name]
 }
 
-async function updateServiceGetVersion (javaServiceFileName, version) {
+async function updateServiceGetVersion(javaServiceFileName, version) {
```
ggrossetie, line 25: Could you please preserve the space before `(` ?

ci/tasks/update-versions.js
```diff
@@ -258,10 +279,12 @@ try {
   await updateServiceGetVersion('Plantuml.java', diagramLibraryVersions.plantuml)
   await updateServiceGetVersion('Structurizr.java', diagramLibraryVersions.structurizr)
   await updateServiceGetVersion('Svgbob.java', diagramLibraryVersions.svgbob)
+  await updateServiceGetVersion('Symbolator.java', diagramLibraryVersions.symbolator)
   await updateServiceGetVersion('Umlet.java', diagramLibraryVersions.umlet)
   await updateServiceGetVersion('Wavedrom.java', diagramLibraryVersions.wavedrom)
   await updateServiceGetVersion('Wireviz.java', diagramLibraryVersions.wireviz)
   await updateVegaServiceGetVersion(diagramLibraryVersions.vega, diagramLibraryVersions.vegalite)
+  await updateSymbolatorAppImageVersion("main.yaml", diagramLibraryVersions.symbolator)
```
ggrossetie, line 287:
```suggestion
```

server/ops/docker/jdk11-jammy/Dockerfile
ggrossetie: That might be because Blockdiag loads "drawer" at runtime using package discovery: https://github.com/blockdiag/blockdiag/blob/ce37f0c18dc7f88caab3d3b24aac4f3f066b793c/src/blockdiag/imagedraw/__init__.py#L23-L31

But maybe we can explicitly tell Nuitka that these resources must be included. Anyway, let's discuss this further in a dedicated issue. I think we can create two issues one for WireViz and one for Blockdiag.

ggrossetie: @zebreus thanks again for mentioning Nuitka! It does a really good job. I was able to build a single binary for blockdiag and WireViz. Now, I just need to implement the "read from stdin/output to stdout" feature :)

ci/tasks/update-versions.js
```diff
@@ -83,7 +83,7 @@ async function mvnEvaluateExpression (expression, pomRootRelativePath) {
   })
 }
 
-function addDiagramLibraryPackageVersion (diagramName, dependencyName, directoryName) {
+function addDiagramLibraryPackageVersion(diagramName, dependencyName, directoryName) {
```
ggrossetie, line 86:
```suggestion
function addDiagramLibraryPackageVersion (diagramName, dependencyName, directoryName) {
```

ci/tasks/update-versions.js
```diff
@@ -56,7 +56,7 @@ async function updateVegaServiceGetVersion (vegaVersion, vegaLiteVersion) {
   await fs.writeFile(servicePath, javaContent, 'utf8')
 }
 
-async function mvnEvaluateExpression (expression, pomRootRelativePath) {
+async function mvnEvaluateExpression(expression, pomRootRelativePath) {
```
ggrossetie, line 59:
```suggestion
async function mvnEvaluateExpression (expression, pomRootRelativePath) {
```

PR #1570, opened 2023-06-12
Add D2 layout engine option

### LINKED ISSUE #1557: Support for alternate layout engine(s) in D2
Currently, D2 diagrams are rendered with the Dagre layout engine. D2 supports multiple different layout engines (among them a commercial one, but that's beside the point and quite probably out of focus for Kroki); there is at least one other FOSS engine available named ELK which may produce superior output depending on what your type of input is. It appears to be immediately usable after installation (with the `-l elk` parameter)

This could be supported via an option just like different themes are supported today.

FelixVanOost:
Adds support for alternate D2 layout engines (see [documentation](https://d2lang.com/tour/layouts/)). Currently this includes both the default layout engine, `dagre`, and an open-source third-party alternative, `ELK`.

The `TALA` engine is not supported because it is proprietary (non-free) and closed-source.

Output with no layout specified (`dagre` is used by default):

```
curl http://localhost:8000/d2/svg/eNo9zLkRgDAMRNF8q3ADtEAvvrExBvpPkH9A8rSjkdZvuwuKjpHMrILBLFhV_3xgw6RuRvRsupp50jM06LnIy6nJbyIXXF-3Hm6y-X5cSiFx > default.svg
```

default

Output with `layout=elk`:

```
curl http://localhost:8000/d2/svg/eNo9zLkRgDAMRNF8q3ADtEAvvrExBvpPkH9A8rSjkdZvuwuKjpHMrILBLFhV_3xgw6RuRvRsupp50jM06LnIy6nJbyIXXF-3Hm6y-X5cSiFx?layout=elk > elk.svg
```

elk

Closes #1557

### DIFF
--- a/server/src/main/java/io/kroki/server/service/D2.java
+++ b/server/src/main/java/io/kroki/server/service/D2.java
@@ -89,8 +89,8 @@ public class D2 implements DiagramService {
   private byte[] d2(byte[] source, JsonObject options) throws IOException, InterruptedException, IllegalStateException {
     List<String> commands = new ArrayList<>();
     commands.add(binPath);
-    String value = options.getString("theme");
-    if (value != null) {
+    String theme = options.getString("theme");
+    if (theme != null) {
       int themeId = 0;
       Integer builtinThemeId = builtinThemes.get(value.toLowerCase().replaceAll("\\s", "-"));
       if (builtinThemeId != null) {
@@ -103,6 +103,11 @@ public class D2 implements DiagramService {
         }
       }
       commands.add("--theme=" + themeId);
+    }
+    String layout = options.getString("layout");
+    // Only pass the layout argument if the ELK layout engine is requested (default is 'dagre')
+    if (layout == "elk") { 
+      commands.add("--layout=" + layout);
     }
     String sketch = options.getString("sketch");
     if (sketch != null) {
--- a/server/src/main/java/io/kroki/server/service/D2.java
+++ b/server/src/main/java/io/kroki/server/service/D2.java
@@ -92,12 +92,12 @@ public class D2 implements DiagramService {
     String theme = options.getString("theme");
     if (theme != null) {
       int themeId = 0;
-      Integer builtinThemeId = builtinThemes.get(value.toLowerCase().replaceAll("\\s", "-"));
+      Integer builtinThemeId = builtinThemes.get(theme.toLowerCase().replaceAll("\\s", "-"));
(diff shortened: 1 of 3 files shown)


ggrossetie:
It looks good to me.
@FelixVanOost Is it ready for review?

FelixVanOost:
@ggrossetie I made these changes while travelling and haven't had time to test them yet. If someone wants to test that the `ELK` layout engine works, that should be enough. Otherwise I can do it myself next week!

f3l1x:
Looking forward. 👍

FelixVanOost:
@ggrossetie I finally had time to test this; results are in the PR description above. It's working!

ggrossetie:
That's quite nice 😍

server/src/main/java/io/kroki/server/service/D2.java
```diff
@@ -89,21 +89,26 @@ public void convert(String sourceDecoded, String serviceName, FileFormat fileFor
...
-          themeId = Integer.parseInt(value, 10);
+          themeId = Integer.parseInt(theme, 10);
         } catch (NumberFormatException e) {
           // ignore, fallback to 0
         }
       }
       commands.add("--theme=" + themeId);
     }
+    String layout = options.getString("layout");
+    if (layout != null && layout.equals("elk")) {
```
ggrossetie, line 108: (nitpick/suggestion): can be simplified to:

```suggestion
    if ("elk".equals(layout)) {
```

or:

```suggestion
    if (Objects.equals(layout, "elk")) {
```

But I'm fine with your code.

FelixVanOost: I initially wrote it like your first suggestion, but I was getting a null pointer error when `layout` wasn't defined. I'm not actually that familiar with Java but I assume that this is because the `equals()` function doesn't check for null pointers before dereferencing. It only started working properly once I added the explicit check for null values first. 🤷🏻‍♂️

ggrossetie: > I initially wrote it like your first suggestion, but I was getting a NullPointerException when layout wasn't defined.

You need to call `equals` on a non-null object:


```java
if ("elk".equals(layout)) { } // OK
```

```java
if (layout.equals("elk")) { } // NullPointerException if layout is null
```

Arguably, the first option does not read well since we want to know if layout is equals to "elk" (the opposite sounds odd "if elk is equals to layout").

Anyway, I'm fine with it 😉

PR #1585, opened 2023-06-23
feat: added C# example to Encode Diagrams

woutervanranst:


docs/modules/setup/pages/encode-diagram.adoc
```diff
@@ -161,3 +162,8 @@ Also, see the Tcl package {uri-dia2kroki}[dia2kroki] which has procedures to bot
 ----
 include::example$encode.ex[]
 ----
+
+[#csharp]
+== C#
+
+See https://github.com/yuzutech/kroki/issues/1584#issuecomment-1604189160.
```
ggrossetie, line 169: We should create a file named `encode.cs` and include it in the documentation:

```
[,csharp]
----
include::example$encode.cs[]
----
```

We could also mention that we cannot use directly https://learn.microsoft.com/fr-fr/dotnet/api/system.io.compression.deflatestream?view=net-7.0 because it does not include  ZLIB header/trailer.

PR #1649, opened 2023-09-27
track blocked thread event and report statistics in metrics endpoint

khanguyen88:
Vert.x monitors and reports blocked thread. This PR will include that info in the /healthcheck endpoint so we can use that info to restart Kroki when perfomance degrades too much due to many blocked threads.

### DIFF
--- a/server/src/main/java/io/kroki/server/service/HealthHandler.java
+++ b/server/src/main/java/io/kroki/server/service/HealthHandler.java
@@ -17,8 +17,6 @@ public class HealthHandler {
   private final String krokiBuildHash;
   private final List<ServiceVersion> serviceVersions;
 
-  private final KrokiBlockedThreadChecker blockedThreadChecker;
-
   public HealthHandler(Map<String, String> versions) {
     this(versions, null);
   }
@@ -31,7 +29,6 @@ public class HealthHandler {
     for (Map.Entry<String, String> entry : versions.entrySet()) {
       serviceVersions.add(new ServiceVersion(entry.getKey(), entry.getValue()));
     }
-    this.blockedThreadChecker = blockedThreadChecker;
   }
 
   public Handler<RoutingContext> create() {
@@ -46,10 +43,6 @@ public class HealthHandler {
       data.put("version", versions);
       for (ServiceVersion serviceVersion : serviceVersions) {
         versions.put(serviceVersion.getService(), serviceVersion.getVersion());
-      }
-      if (blockedThreadChecker != null) {
-        data.put("blockedWorkerPercentage", blockedThreadChecker.blockedWorkerThreadPercentage());
-        data.put("blockedEventLoopPercentage", blockedThreadChecker.blockedEventLoopThreadPercentage());
       }
       routingContext
         .response()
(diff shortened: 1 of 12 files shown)


ggrossetie:
I think that's a good idea.
Did you investigate to understand why threads are blocked? It would be better to fix the root cause rather than periodically restart the service.

khanguyen88:
I had quite a few outages due to blocked threads. So far the logs indicated that certain inputs will cause PlantUml/Ditaa to hang or go into infinite loop. I'm looking out for those bad inputs, but in the meantime, I'm adding as the first step to improve reliability of my server.

ggrossetie:
Are you using the latest version? We are now using single executable binaries (produced by GraalVM) for both PlantUML and Ditaa so it shouldn't block threads anymore.

khanguyen88:
yes, I'm on 0.22

server/src/main/java/io/kroki/server/service/KrokiBlockedThreadChecker.java
```diff
@@ -0,0 +1,85 @@
...
+public class KrokiBlockedThreadChecker {
+
+  private static final Logger logger = LoggerFactory.getLogger(KrokiBlockedThreadChecker.class);
+
+  private final int evenLoopPoolSize;
+  private final int workerPoolSize;
+  private final Duration trackStatsFor;
+
+  private final Map<String, Instant> eventLoopStats = new ConcurrentHashMap<>();
+  private final Map<String, Instant> workerStats = new ConcurrentHashMap<>();
```
ggrossetie, line 28: I'm a bit concerned since we don't do eviction since is basically a memory leak. We could use something like https://github.com/ben-manes/caffeine to use a time-based eviction.

server/src/main/java/io/kroki/server/service/HealthHandler.java
```diff
@@ -41,6 +47,10 @@ public Handler<RoutingContext> create() {
       for (ServiceVersion serviceVersion : serviceVersions) {
         versions.put(serviceVersion.getService(), serviceVersion.getVersion());
       }
+      if (blockedThreadChecker != null) {
+        data.put("blockedWorkerPercentage", blockedThreadChecker.blockedWorkerThreadPercentage());
+        data.put("blockedEventLoopPercentage", blockedThreadChecker.blockedEventLoopThreadPercentage());
+      }
```
ggrossetie, line 53: I wonder if we should export data using Prometheus format on `/metrics`.

khanguyen88: I don't use Prometheus on my server but I'm happy to help. Can you create another issue and assign to me? I'll work on it when I have the bandwidth

ggrossetie: Thanks, I've created an issue

server/src/main/java/io/kroki/server/service/KrokiBlockedThreadChecker.java
khanguyen88: Sure, I switched to Caffeine's Cache to store the stats. 

I was reluctant to use Caffeine as it was not included in the project and I didn't expect the 2 maps to grow above 100 entries.

khanguyen88:
Hi @ggrossetie could you please review and merge this PR?

boris-moduscreate:
@ggrossetie Any eta on a release which includes this? ❤️

ggrossetie:
I will try to push a new release before my vacation (mid January)

PR #1682, opened 2023-12-14
fix: add oci labels to other images

MindTooth:
This way we can have them automatically grouped when e.g. using Renovate.

```
org.opencontainers.image.source="https://github.com/yuzutech/kroki"
```
is the important bit.

### DIFF
--- a/bpmn/Dockerfile
+++ b/bpmn/Dockerfile
@@ -1,5 +1,12 @@
 FROM node:18.18-alpine3.18
 
+LABEL \
+    org.opencontainers.image.title="Kroki - BPMN" \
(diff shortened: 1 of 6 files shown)


MindTooth:
However, it might be better to set the common ones in the bake file and only override individual container. 🤔   I'll look into this instead.

MindTooth:
I reworked it to insert them via bake.  I am unable to run `make buildServer` and `make buildDockerImages` as it crashes somehow.

I'm on macOS M1, and Java 21.  So guess something is up. 🤔

ggrossetie:
Thanks for your contribution! Let's see how it goes in the CI 😄

PR #1685, opened 2024-01-02
Add option for max header size

mattvaughan:
I've deployed Kroki to a setup where the XFCC header size is larger than the 8kb limit. This let's the user set their desired max header size.


ggrossetie:
Thanks!
It looks good to me.

PR #1689, opened 2024-01-10
Update Docker image to Java 17

### LINKED ISSUE #1688: Update Java version to 17 or 21
Hi, would an update of kroki's Java version to 17 or 21 be welcome?

The company I work at uses kroki internally and has a policy for only allowing the last two LTS versions in production and would thus like to update to Java 17 or later.

If it's welcome, I'd open a PR to do the update.

heyLu:
This updates the Java version Kroki uses to Java 17.

For now this is a draft as I found it a bit difficult to test Kroki locally.  (I have IPv6 disabled which some tests seem to require and at least one of the Vega tests fails locally because of some Node.js stuff not being in place as expected.  And even with a hack to get Node.js working it seems to generate a different SVG on my machine.)

Fixes #1688 when merged.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -189,10 +189,10 @@ jobs:
     - name: Install GraphViz
       run: sudo apt-get install graphviz -y
 
-    - name: Set up JDK 11
+    - name: Set up JDK 17
       uses: actions/setup-java@v4
       with:
-        java-version: 11
+        java-version: 17
         distribution: temurin
 
     - name: Build Java server
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,10 +12,10 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install GraphViz
         run: sudo apt-get install graphviz -y
-      - name: Set up JDK 11
+      - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
-          java-version: 11
+          java-version: 17
           distribution: temurin
       - name: Build Java server
         run: make buildServer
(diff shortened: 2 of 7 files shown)


ggrossetie:
Maybe we should do an intermediate step where the minimum version is Java 11 but we use Java 17 as base image? That way we can give a heads-up in the next release for users running Kroki from the jar file. What do you think?

heyLu:
> Maybe we should do an intermediate step where the minimum version is Java 11 but we use Java 17 as base image? That way we can give a heads-up in the next release for users running Kroki from the jar file. What do you think?

@ggrossetie, sounds good, done!  I changed the PR title to reflect this as well.

From my side this is good to merge now.  I can't run the tests locally though¹ (neither on this branch nor on `main`), can you trigger the CI tests to run on this branch?  On the previous runs it looked like only the "test-container" checks ran, not the other CI pipelines.  (Or just merge if things look fine to you.)

¹ A few tests generate differing SVG output with slightly different floating point results.  I suspect somehow my versions are not matching up, but I couldn't figure out what's different.  I am running `node 18.18.2`, which seems to be what generates the differences in some tests.  But as this happens on `main` as well I suspect it's just my setup that is different.

ggrossetie:
> ¹ A few tests generate differing SVG output with slightly different floating point results. I suspect somehow my versions are not matching up, but I couldn't figure out what's different. I am running node 18.18.2, which seems to be what generates the differences in some tests. But as this happens on main as well I suspect it's just my setup that is different.

It can be related to your operating system and/or fonts. I will run the tests on the CI.

ggrossetie:
Thanks!

PR #1705, opened 2024-02-09
Add support for generating Structurizr legend.

### LINKED ISSUE #1686: Structurizr: How to generate keys / legends?
Is there a way to generate the keys / legends associated with a Structurizr view?

It seems that if you're using the Structurizr CLI to export the plantUML views, the legends are generated as separate files, as e.g. `structurizr-systemContext-key.puml` for the key for the systemContext view.

However, providing systemContext-key to Kroki doesn't generate the legend.

willson556:
Add `get-legend` flag that causes Kroki to return the diagram's legend instead of the diagram itself.

I initially attempted to add `include-legend` flag and then to concatenate the UML of the diagram and legend (as discussed in #1686). However, the layout of this was not optimal, at least in my test diagram, the legend was split on either side of the diagram. This solution enables the caller to compose the generated diagram and legend in any way they see fit.

Resolves #1686.


willson556:
> Thanks!
Thank you for the quick review!

> Would you be able to add a test in: https://github.com/yuzutech/kroki/blob/main/server/src/test/java/io/kroki/server/service/StructurizrServiceTest.java ? You can also add a smoke test here: https://github.com/yuzutech/kroki/blob/main/ci/tests/smoke.js.

sure, no problem

> > Add get-legend flag that causes Kroki to return the diagram's legend instead of the diagram itself.
> 
> I don't have a strong opinion but it might be more forward-looking to use `output=legend` (default to `diagram`). If one day there's another output, we could add it and we could also support `output=legend,diagram` if we find a way to combine them? What do you think?

I agree, that's a good idea, I'l make that change as well.

willson556:
@ggrossetie I think everything is ready for another review. Thanks!

ggrossetie:
Looking good, thanks 🙌🏻

PR #1724, opened 2024-03-15
chore(docs): add hugomods-kroki to list of 3rd-party tools

razonyang:


ggrossetie:
Hey @razonyang would it be possible to add a configuration to use a self-hosted installation of Kroki? I dont mind if, by default, a tool is using the public instance but it's important to support self-hosted installations.

razonyang:
Yes, there is [`kroki.endpoint`](https://github.com/hugomods/kroki/blob/bf731b0f653347c8fbd2565ed1572e8bdbd9be4b/hugo.toml#L2) parameter, which is not documented yet.

ggrossetie:
Perfect, thanks!

Issue #1760, opened 2024-07-18
Increase body limit when rendering diagrams


eryalito:
Hey!

I found this https://github.com/yuzutech/kroki/issues/1332 when rendering a small but full of images excalidraw diagram. I didn't found a way of changing this limit without some code changes. Seems that the tweak on this would be minimal, maintaining backwards compatibility by just changing the line 

https://github.com/yuzutech/kroki/blob/5023863bfd81f58d46b9947e8b354d97b5352370/excalidraw/src/index.js#L14

to something like

```js
const diagramSource = await micro.text(req, { limit: (process.env.MAX_BODY_SIZE ?? '1mb'), encoding: 'utf8' })
```

I think a similar approach would be fine in other tools:
https://github.com/yuzutech/kroki/blob/main/diagrams.net/src/index.js#L22
https://github.com/yuzutech/kroki/blob/main/mermaid/src/index.js#L30
...

Would you accept a PR including these changes?

ggrossetie:
Yes sure, the environment variable must be prefixed by `KROKI_` but apart from that 👍🏻

ggrossetie:
Fixed in https://github.com/yuzutech/kroki/commit/21c486c4df824ef7d2c0f7bdd14c63e82b6ce29f

PR #1761, opened 2024-07-20
Change max body size using envvars when rendering diagramas

eryalito:
From #1760 

Allow changing max body sizes using environment variables on mermaid, diagrams.net and excalidraw.

Would it be better to use three different variables, one for each tool (e.g. KROKI_DIAGRAMSNET_MAX_BODY_SIZE) ?


ggrossetie:
Thanks, it looks good to me!
It would be great if you could update this page https://docs.kroki.io/kroki/setup/configuration/ to mention that a new  environment variable is available to update the max body size.

PR #1762, opened 2024-07-22
Add KROKI_MAX_BODY_SIZE variable documentation

eryalito:
Add documentation about new environment variable as suggested on https://github.com/yuzutech/kroki/pull/1761#issuecomment-2242604257


ggrossetie:
Thanks!!
Could you please mention that this environment variable is only available on the mermaid, excalidraw and diagramsnet containers? Otherwise, users might assume that this environment variable can be defined on the gateway/main container (i.e., `yuzutech/kroki`)

eryalito:
Fixed, I hadn't thought about it. I added it as a note, if that's ok

ggrossetie:
Perfeeeect! Many thanks 🙌🏻

PR #1767, opened 2024-08-10
Structurizr DSL: set restricted parsing

NicolaiSoeborg:
Enable restricted parsing to disable `!script` and other dangerous methods to be executed during parsing

### DIFF
--- a/server/src/main/java/io/kroki/server/service/Structurizr.java
+++ b/server/src/main/java/io/kroki/server/service/Structurizr.java
@@ -88,6 +88,7 @@ public class Structurizr implements DiagramService {
   static byte[] convert(String source, FileFormat fileFormat, PlantumlCommand plantumlCommand, StructurizrPlantUMLExporter structurizrPlantUMLExporter, JsonObject options) throws IOException, InterruptedException {
     StructurizrDslParser parser = new StructurizrDslParser();
     try {
+      parser.setRestricted(true);
       parser.parse(source);
       ViewSet viewSet = parser.getWorkspace().getViews();
       Collection<View> views = viewSet.getViews();
--- a/server/src/main/java/io/kroki/server/service/Structurizr.java
+++ b/server/src/main/java/io/kroki/server/service/Structurizr.java
@@ -10,6 +10,7 @@ import io.kroki.server.decode.SourceDecoder;
 import io.kroki.server.error.BadRequestException;
 import io.kroki.server.error.DecodeException;
 import io.kroki.server.format.FileFormat;
+import io.kroki.server.security.SafeMode;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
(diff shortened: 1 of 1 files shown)


ggrossetie:
Thanks!
I think we should introduce a configuration (i.e., environment variable) to configure this value. Some people might want to disable the restricted parsing on their private deployment.

```
KROKI_STRUCTURIZR_RESTRICTED_PARSER
```

The default value should be `true` (secured by default).
You can take a look at the PlantUML service where we use the config object:

https://github.com/yuzutech/kroki/blob/ba7a512be669697b521aae47a64a5d2f79cea34a/server/src/main/java/io/kroki/server/service/Plantuml.java#L117

server/src/main/java/io/kroki/server/service/Structurizr.java
```diff
@@ -48,6 +50,7 @@ public class Structurizr implements DiagramService {
 
   public Structurizr(Vertx vertx, JsonObject config) {
     this.vertx = vertx;
+    this.safeMode = SafeMode.get(config.getString("KROKI_STRUCTURIZR_RESTRICTED_PARSER", "secure"), SafeMode.SECURE);
```
ggrossetie, line 53: To be honest, I hadn't thought about reusing the concept of safe mode. I had envisioned a configuration that would restrict the parser with a boolean value. If we want to reuse the concept of safe mode, I suppose we should do the following:

- Rename the environment variable to `KROKI_STRUCTURIZR_SAFE_MODE`
- Use the value of `KROKI_SAFE_MODE` if `KROKI_STRUCTURIZR_SAFE_MODE` is not defined
- Explain this functionality in the documentation: https://github.com/yuzutech/kroki/blob/358075666a392e1f58885f6340344b761b5a037c/docs/modules/setup/pages/configuration.adoc?plain=1#L47


What do you think?

ggrossetie:
Build failure:
```
kroki/server/src/main/java/io/kroki/server/service/Structurizr.java:[94,28] non-static variable this cannot be referenced from a static context
```

Since the `convert` method is static, you cannot reference `this.safeMode`: 

https://github.com/yuzutech/kroki/blob/358075666a392e1f58885f6340344b761b5a037c/server/src/main/java/io/kroki/server/service/Structurizr.java#L88

You will most likely need to pass the safe mode when calling `convert`:

https://github.com/yuzutech/kroki/blob/358075666a392e1f58885f6340344b761b5a037c/server/src/main/java/io/kroki/server/service/Structurizr.java#L80

NicolaiSoeborg:
I kinda like a "global safeMode" concept (maybe other formats will also benefit in case they allow 'unsafe' parsing?)

Initially I quickly edited using the github editor, so didn't go too deep into the code - thanks for the detailed analysis!

Feel like changing some instances of parameters to `convert` might be a bit error-prone? (and we are constrained by the `DiagramService` interface - maybe we can store it in `options`?)

I thought this would be a quick one-liner :upside_down_face:

ggrossetie:
Superseded by https://github.com/yuzutech/kroki/pull/1767

PR #1770, opened 2024-08-12
Replace dvisvgm builder with apt package

felixvanoost:
- Replaces the `kroki-builder-dvisvgm` Docker builder image with the now available apt package. `dvisvgm` is available from Ubuntu 24.04 onwards.
- Updates the base and builder images to use Ubuntu 24.04.


ggrossetie:
> 7.353 info: Selecting GID from range 1000 to 59999 ...
> 7.355 info: Adding group `kroki' (GID 1001) ...
> 7.393 fatal: The UID 1000 is already in use.

I think this is a mistake in the Dockerfile as we should use `1001` GID and not `1000`. I guess the latest version of `adduser` is more strict and fail if the GID is already taken.

ggrossetie:
Not sure why but `symbolator` is failing 🤔

ggrossetie:
Still failing, my guess is that symbolator relies on a system dependency that was removed/updated. Do you reproduce this issue locally?

felixvanoost:
> Still failing, my guess is that symbolator relies on a system dependency that was removed/updated. Do you reproduce this issue locally?

I've been having issues running the project locally, hence my over-reliance on the CI here. 🙃

Symbolator was failing because I forgot to update the builder image to Ubuntu 24.04, so it was compiled for Python 3.11 instead of 3.12.

ggrossetie:
Perfect, thanks!

PR #1771, opened 2024-08-13
Drawio - Add resolution of external image to get self contained SVG

johnvox:
This PR help to keep drawio svg self contained.

Background
I use SelfHosted Kroki with antora doc generator
When using drawio diagrams it happned that some part of the diagr was missing because it relies on external images.

What have been done:

With this PR, during the rendering phase, kroki will retreive external image to embed them  as base64 src.
This way diagrram will be rendered correctly whenever and wherever

### DIFF
--- a/diagrams.net/src/worker.js
+++ b/diagrams.net/src/worker.js
@@ -2,7 +2,6 @@
 import path from 'node:path'
 import { URL, fileURLToPath } from 'node:url'
 import puppeteer from 'puppeteer'
-
 import { logger } from './logger.js'
 
 const __dirname = fileURLToPath(new URL('.', import.meta.url))
@@ -15,8 +14,10 @@ export class TimeoutError extends Error {
 
 export class SyntaxError extends Error {
   constructor (err) {
-    logger.error({ err })
-    super(`Syntax error in graph: ${JSON.stringify(err)}`)
+    super('Syntax error in graph', { cause: err })
+    logger.error(this)
+    this.name = 'SyntaxError'
+    this.message = err.message
   }
 }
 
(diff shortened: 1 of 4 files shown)


diagrams.net/src/worker.js
```diff
@@ -27,6 +29,32 @@ export class Worker {
...
+      try {
+        let imgContent = await readFile(filePath)
+        nodeElement.attr("xlink:href",`data:${mimeType};base64,${imgContent.toString('base64')}`)
+        nodeElement.removeAttr("pointer-events")
+      } catch(err){
+        continue
+      }
+    }
+    await writeFile("before.svg", svgRoot)
+    await writeFile("after.svg", $("svg").toString())
```
ggrossetie, line 54: I guess this was for testing purpose?

diagrams.net/src/worker.js
```diff
@@ -27,6 +29,32 @@ export class Worker {
     this.convertTimeout = process.env.KROKI_DIAGRAMSNET_CONVERT_TIMEOUT || '15000'
   }
 
+  async resolveImage(svgRoot) {
+    let $ = cheerio.load(svgRoot,  {
```
ggrossetie, line 33: Since we are using Puppeteer we could potentially parse/edit the SVG in `page.evaluate`? That would remove the dependency on `cheerio`.

jynolen: Definitely I am not a JS expert I will check to change it

diagrams.net/src/worker.js
jynolen: Yes I will remove it

ggrossetie:
Someone mentioned in the community chat that diagrams.net has an "embed" feature:

https://www.drawio.com/doc/faq/embed-mode

It might be possible to rely on the query parameter `embed=1` to enable it (or find a way to set this flag).

ggrossetie:
@jynolen could you please share an example with external images? I want to test if `embed=1` is working as intended.

jynolen:
Here you are 👍 https://gist.github.com/jynolen/05f48d96f09cfdffd9bcf6b99b91256f
I put the two (with url and with b64)

ggrossetie:
It's not as easy as I thought since it relies on a proxy: https://github.com/yuzutech/kroki/issues/1796
So your strategy is probably the best.

As this adds an attack vector, I think it would be preferable to enable this feature when Kroki safe mode is set to `unsafe`: https://docs.kroki.io/kroki/setup/configuration/#_safe_mode

diagrams.net/src/worker.js
ggrossetie: Did you find a way to implement this feature without relying on cheerio? I can take a look if you want?

jynolen:
I update the code to discard unsage of cheerio.
Don't know yet how to implement the unsafe flag

jynolen:
Hello any news from your side ?

jynolen:
Hello, do you think changes can be merged ?

ggrossetie:
@jynolen sorry for the late reply! It looks good 👍🏻 
I will do a more in-depth code review this afternoon before merging

PR #1774, opened 2024-08-20
encode diagram: Google Apps Script example

wivaku:


wivaku:
Please ignore. Google uses gzip (or zip), which is not compatible with Kroki's zlib.

Alternative is to manually copy/paste pako.min.js and then use:
```
function pakoEncode(inputString) {
  var compressedBytes = pako.deflate(inputString, { to: 'string' });
  var base64UrlSafeString = Utilities.base64EncodeWebSafe(compressedBytes);

  return base64UrlSafeString;
}
```

PR #1811, opened 2024-11-26
chore: diagrams.net - add /health API for health checks 

quytm:
Hi team,

I found that `diagrams.net` is missing `/health` for the health check. That may lead to a failure in K8s when starting the instance, so I created this pull request to support it.

Please review it and give me comments if any!

Thank you!

### DIFF
--- a/diagrams.net/src/index.js
+++ b/diagrams.net/src/index.js
@@ -14,6 +14,14 @@ import { SyntaxError, TimeoutError, Worker } from './worker.js'
   const worker = new Worker(browser)
   const server = new http.Server(
     micro.serve(async (req, res) => {
+      // Add a /health route that renders a sample diagram by calling the worker
+      if (req.url === '/health') {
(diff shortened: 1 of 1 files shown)


ggrossetie:
Thanks @quytm !
We still need to work on https://github.com/yuzutech/kroki/issues/1423 but I think it's fine to add the `/health` endpoint since we already have it in other services.

diagrams.net/src/index.js
```diff
@@ -14,6 +14,14 @@ import { SyntaxError, TimeoutError, Worker } from './worker.js'
   const worker = new Worker(browser)
   const server = new http.Server(
     micro.serve(async (req, res) => {
+      // Add a /health route that renders a sample diagram by calling the worker
+      if (req.url === '/health') {
+        let sample = `<mxGraphModel><root><mxCell id="0"/><mxCell id="1" parent="0"/></root></mxGraphModel>`
```
ggrossetie, line 19: I guess it can a `const`?

```suggestion
        const sample = `<mxGraphModel><root><mxCell id="0"/><mxCell id="1" parent="0"/></root></mxGraphModel>`
```

PR #1816, opened 2025-01-13
Update to PlantUML 1.2025.0

diguage:
PlantUML 1.2025.0 support a great feature: https://github.com/plantuml/plantuml/pull/2001


ggrossetie:
We need to build it from https://github.com/yuzutech/plantuml then we can upgrade the version

PR #1828, opened 2025-02-13
fix(plantuml): pass plantuml.include.path property to native plantuml binary

### LINKED ISSUE #1765: Problem with plantuml include and KROKI_PLANTUML_INCLUDE_PATH
Hi,

I have local installation of kroki with plantuml libs stored in /home/kroki/include

Kroki is run with the following settings:

```
  - name: KROKI_PLANTUML_ALLOW_INCLUDE
    value: "true"
  - name: KROKI_PLANTUML_INCLUDE_PATH
    value: "/home/kroki/include"
  - name: KROKI_SAFE_MODE
    value: "SAFE"   
```                                                         

In kroki versions up to 0.20.0 the following works fine:

```
@startuml JWTInjection
!global $LIB_BASE_LOCATION="tmorin-plantuml-libs-11.1.0"
!include $LIB_BASE_LOCATION/bootstrap.puml   
```                                                               

However, since 0.21.0, I get:
`"Error 400: cannot include tmorin-plantuml-libs-11.1.0/bootstrap.puml"`
and, as a workaround, I need to use full path:

```
@startuml JWTInjection
!global $LIB_BASE_LOCATION="/home/kroki/include/tmorin-plantuml-libs-11.1.0"
!include $LIB_BASE_LOCATION/bootstrap.puml
```

What happened to KROKI_PLANTUML_INCLUDE_PATH in recent versions of kroki?

thanks,
Michal

mlyczek:
I have found a way of passing the `plantuml.include.path` property to the native plantuml library.

It's worth noting that, plantuml has a `-D` parameter which sets value of global preprocessing variable as described here: https://plantuml.com/command-line, however when built using GraalVM into native library, the GraalVM handling of the `-D` parameter (described here: https://www.graalvm.org/latest/reference-manual/native-image/guides/use-system-properties/) takes precedence, which is the reason why this fix works.

I have added debug log message to add ability to see the exact plantuml command parameters which makes it easier to debug any problems with it, if necessary.

Please let me know what do you think.

Fixes #1765

### DIFF
--- a/server/src/main/java/io/kroki/server/service/Plantuml.java
+++ b/server/src/main/java/io/kroki/server/service/Plantuml.java
@@ -125,11 +125,6 @@ public class Plantuml implements DiagramService {
     this.ditaaCommand = new DitaaCommand(config);
     this.includeWhitelist = parseIncludeWhitelist(config);
     this.logging = new Logging(logger);
-    // Disable unsafe include for security reasons
-    String plantUmlIncludePath = config.getString("KROKI_PLANTUML_INCLUDE_PATH");
-    if (plantUmlIncludePath != null) {
-      System.setProperty("plantuml.include.path", plantUmlIncludePath);
-    }
   }
 
   static List<Pattern> parseIncludeWhitelist(JsonObject config) {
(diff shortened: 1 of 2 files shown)


ggrossetie:
Thanks a lot @mlyczek 
It looks good 👍🏻

PR #1830, opened 2025-02-15
chore: upgrade PlantUML to 1.2025.0

ggrossetie:


ggrossetie:
Down to one test failing.
Somehow, PlantUML (native image) is unable to load/decode images from https, for instance:

```
Cannot decode https://static.structurizr.com/themes/amazon-web-services-2020.04.30/AWS-Cloud_light-bg@4x.png
```

PR #1832, opened 2025-02-19
Update Taskfile.yml - Fix windows build using mavenBuild

lyoneel:
Fix windows build using mavenBuild, using the correct script for windows mvnw.cmd


ggrossetie:
Thanks 👍🏻

PR #1854, opened 2025-03-12
chore(deps): update Docker base images

gustavclausen:
To solve [CVE-2025-26519](https://nvd.nist.gov/vuln/detail/CVE-2025-26519) amongst other security vulnerabilities.

### DIFF
--- a/bpmn/Dockerfile
+++ b/bpmn/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:lts-alpine3.21
+FROM node:22.14-alpine3.21
 
 RUN addgroup -g 1001 kroki && adduser -D -G kroki -u 1001 kroki
 
--- a/bpmn/Dockerfile
+++ b/bpmn/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:20.18-alpine3.20
+FROM node:lts-alpine3.21
 
 RUN addgroup -g 1001 kroki && adduser -D -G kroki -u 1001 kroki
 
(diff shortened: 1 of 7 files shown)


gustavclausen:
@ggrossetie Would it be possible to update to latest version (23-alpine3.21) while we are at it?

bpmn/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:20.18-alpine3.19
```
ggrossetie, line 1: I would prefer to use an explicit tag otherwise we will switch to the next LTS (when we don't expect it).

gustavclausen: Makes total sense 👍 I just understood that you wanted to target the `lts` tag.

Do you want to target only the major version (e.g. `22-alpine3.21`) or including the minor version like it is done currently (e.g. `22.14-alpine3.21`)?

ggrossetie: > Makes total sense 👍 I just understood that you wanted to target the lts tag.

Sorry, I wasn't clear 😉 

> Do you want to target only the major version (e.g. 22-alpine3.21) or including the minor version like it is done currently (e.g. 22.14-alpine3.21)?

I think major + minor is good so `22.14-alpine3.21`

gustavclausen: Great, fixed now

gustavclausen: @ggrossetie Could I request a re-review?

ggrossetie: it looks good, thanks 👍🏻

Issue #1874, opened 2025-04-16
Kroki container fails to draw a blockdiag diagram


alex-mashin:
Kroki is installed with the following fragment of `docker-compose.yml`:

```
  kroki:
    container_name: kroki
    image: yuzutech/kroki
    depends_on:
      - kroki_mermaid
      - kroki_bpmn
      - kroki_excalidraw
    environment:
      - KROKI_MERMAID_HOST=kroki_mermaid
      - KROKI_BPMN_HOST=kroki_bpmn
      - KROKI_EXCALIDRAW_HOST=kroki_excalidraw    
    restart: unless-stopped
    tmpfs:
      - /tmp
    networks:
      kroki:
    cpu_shares: 1024
    deploy:
      resources:
        limits:
          memory: 512m

(kroki_mermaid, kroki_bpmn and kroki_excalidraw are skipped)
```

kroki is called from another container with a POST request to `http://kroki:8000/blockdiag/svg`.

The body of the request is

```
blockdiag {
  Kroki -> generates -> "Block diagrams";
  Kroki -> is -> "very easy!";

  Kroki [color = "greenyellow"];
  "Block diagrams" [color = "pink"];
  "very easy!" [color = "orange"];
}
```

The responce is:

```
Error 400: Error, couldn't launch child (exec): Permission denied
 (exit code 2)
```

The container log contains:

```
{"timestamp":"1744815480814","level":"INFO","thread":"vert.x-eventloop-thread-1","mdc":{"path":"/blockdiag/svg","method":"POST","service_name":"blockdiag","action":"request_received","bytes_read":"192","user_agent":"MediaWiki/1.39.11"},"logger":"io.kroki.server.service.DiagramHandler","message":"Request received POST /blockdiag/svg","context":"default"}
{"timestamp":"1744815481073","level":"INFO","thread":"vert.x-eventloop-thread-1","mdc":{"path":"/blockdiag/svg","took":"256","method":"POST","service_name":"blockdiag","action":"convert","file_format":"svg","user_agent":"MediaWiki/1.39.11"},"logger":"io.kroki.server.service.DiagramHandler","message":"Convert took 256ms","context":"default"}
[...]

PR #1875, opened 2025-04-18
Document /tmp:exec

alex-mashin:
Document the fact that, if /tmp is explicitly mounted into kroki container, the option exec: ought to be used.

Fixes: #1874


docs/modules/setup/examples/kroki-docker-compose.yml
```diff
@@ -11,6 +11,9 @@ services:
       - KROKI_EXCALIDRAW_HOST=excalidraw
     ports:
       - "8000:8000"
+    # If you need at all to explicitly mount /tmp, add exec option.
```
ggrossetie, line 14:
```suggestion
    # If you do need to explicitly mount /tmp, make sure to include the exec option
```

PR #1882, opened 2025-04-28
Modernize C# example

RamType0:
- `ZLibStream` has been added in .NET 6.0
- `Base64Url` has been added in .NET 9.0
- Append the checksum-trailer to `MemoryStream` directly


ggrossetie:
Thanks!

PR #1891, opened 2025-05-09
feat: Allow redirect to companion server

### LINKED ISSUE #1887: Kroki server does not accept redirect to companion server.
When we deploy kroki companion server as Azure Container Apps with private endpoints, kroki  main server(which is also deployed as Azure Container Apps) could access companion server via 301 redirect.

But kroki server could not handle redirect.

```json
 {
    "timestamp":"1746709627078",
    "level":"WARN",
    "thread":"vert.x-eventloop-thread-1",
    "logger":"io.kroki.server.error.ErrorHandler",
    "message":"Unexpected error code in ErrorHandler. Got: 301. Error code must be within 400 and 599, fallback to 500",
    "context":"default"
}
```
This makes us impossible to host kroki companion servers on Azure Container Apps.

RamType0:
Fix #1887

### DIFF
--- a/server/src/main/java/io/kroki/server/Server.java
+++ b/server/src/main/java/io/kroki/server/Server.java
@@ -1,6 +1,7 @@
 package io.kroki.server;
 
 import io.kroki.server.action.Commander;
+import io.kroki.server.action.Delegator;
 import io.kroki.server.error.ErrorHandler;
 import io.kroki.server.error.InvalidRequestHandler;
 import io.kroki.server.log.Logging;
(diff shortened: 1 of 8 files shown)


server/src/main/java/io/kroki/server/Server.java
```diff
@@ -72,7 +79,59 @@ static void start(Vertx vertx, VertxOptions vertxOptions, JsonObject config, Han
     HttpServer server = vertx.createHttpServer(serverOptions);
     Router router = Router.router(vertx);
     BodyHandler bodyHandler = BodyHandler.create(false).setBodyLimit(config.getLong("KROKI_BODY_LIMIT", BodyHandler.DEFAULT_BODY_LIMIT));
-
+    HttpClient delegatorHttpClient = vertx.httpClientBuilder().withRedirectHandler(resp -> {
```
ggrossetie, line 82: could you please add a link to the original code and explain why we are not using the default redirectHandler (i.e., why we are not using `setFollowRedirects`):

https://github.com/eclipse-vertx/vert.x/blob/78ca74e30d8e0b7b8950c4c4030b45224c2a7e0a/vertx-core/src/main/java/io/vertx/core/http/impl/DefaultRedirectHandler.java#L25

ggrossetie, line 82: according to the document we need to `setFollowRedirects`:

>  The redirect handler is called when a {@code 3xx} response is received and the request is configured to follow redirects with {@link HttpClientRequest#setFollowRedirects(boolean)}.

RamType0: I wrote the reason at [here](https://github.com/yuzutech/kroki/issues/1887#issuecomment-2867744573), but it should be added in comment as you said.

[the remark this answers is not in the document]
RamType0: With current signature, `Delegator` takes `WebClient` as argument.
But we could assign redirect handler only when constructing `HttpClient`.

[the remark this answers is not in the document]
ggrossetie: You can update the method signature, the httpClient can become an instance property.

ggrossetie: As far as I understand, even when using a custom redirect handler we need to declare `setFollowRedirects (true)` otherwise the redirect handler won't be called.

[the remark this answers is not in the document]
RamType0: Do you mean make Delegator as general class rather than current "static" class?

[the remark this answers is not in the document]
ggrossetie: Yes, I think it now makes more sense to initialize the webclient in the Delegator constructor.

RamType0: Added comment

RamType0: https://vertx.io/docs/vertx-web-client/java/#_handling_30x_redirections

> By default the client follows redirections

Also, I added test and it actually works fine without `setFollowRedirects (true)`.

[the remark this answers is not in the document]
RamType0: Done!

RamType0:
@ggrossetie I have done your requests

ggrossetie:
Perfect, thanks 🎉

Issue #1885, opened 2025-05-06
Open Telemetry support


RamType0:
If Kroki server supports Open Telemetry, it would be helpful for us to investigate problem with companion servers, and forwarded request to kroki server.

[Vert.x officially supports Open Telemetry](https://vertx.io/docs/vertx-opentelemetry/java/).

ggrossetie:
resolved in #1892

PR #1892, opened 2025-05-09
feat: Open Telemetry tracer

RamType0:
#1885


ggrossetie:
I think we need to update the documentation to explain how to use it: https://docs.kroki.io/kroki/setup/configuration/
(We can add link to the [official documentation](https://opentelemetry.io/) but we should at least mention that open telemetry is supported in Kroki)

Issue #1895, opened 2025-05-10
Erd with output format jpeg doesn't work


RamType0:
https://kroki.io/erd/jpeg/eNqLDkgtKs7Pi-XSykvMTeXKSM1MzyjhKodQ2kmZRSUZ8Tn5yYklmfl58ZkpXFzRPlAeUAuQn5xZUslVXJJYksqVnF-aV1JUycUFMVJBS1fXUAGmGgCFAiQX

> 500 Internal Server Error

ggrossetie:
I think that's because we need to use `jpg` and not `jpeg` when calling `erd` binary.

PR #1898, opened 2025-05-10
fix: Erd with output format jpeg doesn't work #1895

RamType0:
This might work, but I could not build kroki docker image in my environment (`sudo task dockerBuildImages` never finish).
Does any one can test it?


ggrossetie:
You can add `jpeg` in `outputFormat`:

https://github.com/yuzutech/kroki/blob/a9f7c5b3bcda76ff35ebf4f937785249b254edbb/ci/tests/smoke.js#L20

It will test that `/erd/jpeg` returns a 200.

PR #1926, opened 2025-07-03
Fix bad tikz render due to bad documentclass

### LINKED ISSUE #1870: [TikZ] `ball color` missing when exporting to SVG
I encountered an issue when trying to use the `ball color` key in a TikZ drawing. The following code snippet from https://tex.stackexchange.com/a/79398/260481 does not render as expected:
```latex
\documentclass[margin=3mm]{standalone}
\usepackage{tikz}
\begin{document}
\begin{tikzpicture}
  \def\eggheight{3cm}
  \path[ball color=orange!60!gray]
  plot[domain=-pi:pi,samples=100]
  ({.78*\eggheight *cos(\x/4 r)*sin(\x r)},{-\eggheight*(cos(\x r))})
  -- cycle;
\end{tikzpicture}
\end{document}
```
The `ball color` key appears to have no effect or does not work correctly. Could you please confirm if this is a known issue or provide guidance on resolving it?

Thank you for your assistance!

rfdonnelly:
The tikz svg output uses dvisvgm.  Dvisvgm requires the input latex to specify the PGF dvisvgm driver by passing the dvisvgm option via the documentclass.

Previous to this change, the kroki tikz output would generate a correct documentclass if the user didn't provide one.  However, if the user provided a documentclass the kroki tikz output would use it as is.  This allowed users to specify an incorrect documentclass and then get undesired output.

Fix by overwriting any user-provided documentclass.  The implication is that users can no longer provide their own documentclass.  However, I couldn't find any convincing usecases for this in the context of diagrams.

This change implements option 1 from
https://github.com/yuzutech/kroki/issues/1870#issuecomment-2822066909. The alternative of surgically fixing up the user-provided documentclass instead of completely overwriting it was also considered but decided against due to complexity.  This alternative could be added later if a convincing usecase presents.

Example input that this change fixes:

```latex
\documentclass[margin=3mm]{standalone}
\usepackage{tikz}
\begin{document}
\begin{tikzpicture}
  \def\eggheight{3cm}
  \path[ball color=orange!60!gray]
  plot[domain=-pi:pi,samples=100]
  ({.78*\eggheight *cos(\x/4 r)*sin(\x r)},{-\eggheight*(cos(\x r))})
  -- cycle;
\end{tikzpicture}
\end{document}
```

Before this change:

image

After this change:

image

Fixes #1870

### DIFF
--- a/tikz/tikz2svg
+++ b/tikz/tikz2svg
@@ -34,7 +34,7 @@ backend_driver() {
             echo
             ;;
         svg)
-            echo dvisvgm,tikz
+            echo dvisvgm
             ;;
     esac
 }
@@ -59,7 +59,7 @@ documentclass_present() {
 
 infer_documentclass() {
     output_format=$1
-    echo "\\documentclass[$(backend_driver $output_format)]{standalone}"
+    echo "\\documentclass[tikz,$(backend_driver $output_format)]{standalone}"
 }
 
 add_documentclass() {
--- a/tikz/tikz2svg
+++ b/tikz/tikz2svg
@@ -62,13 +62,30 @@ infer_documentclass() {
     echo "\\documentclass[$(backend_driver $output_format)]{standalone}"
 }
 
-maybe_add_documentclass() {
+add_documentclass() {
+    documentclass=$1
+    raw_input=$2
+
+    echo $documentclass
+    cat $raw_input
+}
+
+replace_documentclass() {
+    documentclass=$1
+    raw_input=$2
+
+    sed "s/\\documentclass.*\\?}/$documentclass/" $raw_input
+}
+
+fixup_documentclass() {
     output_format=$1
     raw_input=$2
-    if ! documentclass_present $raw_input; then
-        infer_documentclass $output_format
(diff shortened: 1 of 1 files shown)


rfdonnelly:
Hmm.  Looks like the test diagram uses a feature of the standalone documentclass to load the tikz package with a documentclass option.  From section 5.1 of https://mirrors.ibiblio.org/CTAN/macros/latex/contrib/standalone/standalone.pdf:  
> image

The documentclass from the test diagram:
https://github.com/yuzutech/kroki/blob/baefb44d3dfe475285e2625a5e051a2162474ad8/ci/tests/diagrams/periodic-table.tex#L6

Since this PR overwrites the documentclass, this diagram is now effectively missing a `\usepackage{tikz}` command.  I think this can be fixed for all cases by simply changing the generated documentclass as follows:

```diff
- \documentclass[dvisvgm]{standalone}
+ \documentclass[dvisvgm,tikz]{standalone}
```

This works whether the diagram contains a `\usepackage{tikz}` command or not.

ggrossetie:
Many thanks!

PR #2035, opened 2026-04-02
Security: Reflected XSS in HTML error page rendering

tuanaiseo:
## Problem

The error page template inserts `{errorMessage}` and `{stackTrace}` directly into HTML without escaping. Error messages can include untrusted content (for example, external tool stderr/stdout propagated through exceptions), which could execute script in a browser viewing the error page.

**Severity**: `high`
**File**: `server/src/main/resources/web/error.html`

## Solution

HTML-escape all dynamic values before template substitution (error message, stack trace, title, logo placeholders if dynamic). Prefer server-side templating with automatic escaping and only allow explicit safe HTML for tightly controlled fields.

## Changes

- `server/src/main/resources/web/error.html` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced


ggrossetie:
The new variables are not defined. Is this an automated AI pull request?

PR #2129, opened 2026-08-08
fix: block Chromium's calls to Google in the companion browser instance

### LINKED ISSUE #2128: Headless chromium instances from companion docker containers are phoning home (to google)
This happens in the current `0.32.0` version and so far in all previous versions I have checked. The `bpmn` and the `excalidraw` containers are doing this as soon as they are started, the `mermaid` container starts contacting `google` as soon as the main container requests a diagram from it.

`tcpdump` shows `A` and `AAAA` requests for the following DNS names
- `accounts.google.com`
- `android.clients.google.com`
- `clients2.google.com`
- `update.googleapis.com`
- `www.google.com`

and then `HTTPS` connects on port `443` (even some `HTTP` connects on port `80`) to those sites.  

It looks like `puppeteer` cannot completely ***silence*** `chromium` by passing `--disable-*` options on launch.

Hence I suggest to use the following `kroki-confined.yml` (`docker-compose`) file which disables outside (docker) network access for the companion containers entirely (by putting them into a `network` with `internal: true`)

> [!NOTE] 
> - This also prevents the companion containers from fetching remote resources which may be referenced in the diagram source passed to them. I think that `kroki` adds a block list to prevent this anyway, unless some `unsafe` flag has been passed.
> - The `dns_search: ["."]` setting for the main container prevents appending your host's **dns search list** when the main container tries to resolve the (internal) IP-addresses of companion containers. This prevents name conflicts with existing hosts on your network (which may have the same name as a companion container, e. g. `excalidraw`). It also prevents ***fishing*** for known hostnames on your network (which `kroki` does not do, I know.)
> - Using `127.0.0.1:8000:8000` for the main container restricts access to localhost only, e. g. a `reverse-proxy` running on your host which does the `SSL termination`.  

```yaml
services:
  kroki:
    image: yuzutech/kroki:0.32.0
    depends_on:
      - mermaid
      - bpmn
      - excalidraw
    environment:
      - KROKI_MERMAID_HOST=mermaid
      - KROKI_BPMN_HOST=bpmn
      - KROKI_EXCALIDRAW_HOST=excalidraw
    networks:
      - external
      - default
    dns_search: ["."]
    ports:
      - "127.0.0.1:8000:8000"

  mermaid:
    image: yuzutech/kroki-mermaid:0.32.9
    expose:
      - "8002"

  bpmn:
    image: yuzutech/kroki-bpmn:0.32.0
    expose:
      - "8003"

  excalidraw:
    image: yuzutech/kroki-excalidraw:0.32.0
    expose:
      - "8004"

networks:
  external:
    name: "kroki_external"
    driver_opts:
      com.docker.network.bridge.name: "kroki_external"

  default:
    name: "kroki_default"
    internal: true
    driver_opts:
      com.docker.network.bridge.name: "kroki_default"
      com.docker.network.bridge.inhibit_ipv4: 1
```

ggrossetie:
- The shared headless Chromium instance (Mermaid, BPMN, Excalidraw, diagrams.net) reaches out to Google (`update.googleapis.com`, `clients2.google.com`, `android.clients.google.com`, `accounts.google.com`, `www.google.com`) as soon as it launches, via several independent built-in subsystems (GCM device checkin, network-time sync, the component updater, ...)
- Disabling each subsystem individually (`--disable-component-update`, `--disable-domain-reliability`, `--no-pings`) still left some of them contacting Google — confirmed by packet capture (tcpdump/tshark)
- Fixed by blocking DNS resolution for `*.google.com` and `*.googleapis.com` at the browser level via `--host-resolver-rules`, so nothing Chromium does can reach Google regardless of which internal subsystem tries

Fixes #2128


xlsigned:
Wow, that was some rapid response.
Thank you!

ggrossetie:
> Wow, that was some rapid response.
Thank you!

Eh eh! could you please give it a try? I'm a bit torn on the host resolver rules option, particularly for Google Fonts https://fonts.googleapis.com/, but I haven't found a better way to block the calls... 😞

xlsigned:
> Eh eh! could you please give it a try? I'm a bit torn on the host resolver rules option, particularly for Google Fonts https://fonts.googleapis.com/, but I haven't found a better way to block the calls... 😞

I can try later tonight. Need to build/update the docker containers myself, right?
Though I'm not sure or rather I doubt, that I have any diagram sources which rely on fonts being downloadable from `fonts.googleapis.com`. Otherwise I would have noticed this already (see below).  

I'm currently using the workaround with a separate (`internal: true`) docker network for the companion containers. In fact, none of the containers has ever been able to actually reach google (or rather anywhere outside) from my host, since I'm running docker with `iptables: false`. The issue actually popped up at our firewall which complained about spoofed IP addresses, since my containers were using their bridge IP as the source address to ***phone home***.

Meanwhile I've learned a lot more about docker networking and I'm running my own set of `iptables` rules which blocks all traffic from the docker bridges (unless I explicitly allow certain IPs and manually add a `MASQUERADE` and a `FORWARD` rule). I still do not like docker to mess with the `iptables/nftables` on my host.

xlsigned:
>  I'm a bit torn on the host resolver rules option, particularly for Google Fonts https://fonts.googleapis.com/ ...

You could always map `update.googleapis.com` to `localhost` and `*.googleapis.com` only when `KROKI_SAFE_MODE == secure` (the default) or `KROKI_SAFE_MODE != unsafe`, right?

Also, if people really want to operate `kroki` in `unsafe` mode, i.e. allow containers to fetch external resources, they probably (should) have their own `iptables/nftables` rules in place. Either on top of the `docker` rules using the `DOCKER_USER` chain or with `iptables[6]: false` using the default chains.

Basically:
- Improve the default `safe/secure` mode, by adding the `--host-resolver-rules` to map the `google` domains to `localhost`.
- Strongly suggest in the documentation to add `iptables/nftables` rules to only allow access to specific sites, if running in `unsafe` mode, i.e. deny all by default (which includes google) and whitelist exactly what you need.

xlsigned:
> > Eh eh! could you please give it a try?
> 
> I can try later tonight. Need to build/update the docker containers myself, right?

@ggrossetie

I've copied the patched `index.js` from the PR to `/usr/local/lib/browser-instance/` of the `bpmn`, `mermaid` and `excalidraw` containers and committed the change to new/fixed images for the companion containers.

When I run a standard docker-compose, i.e. without the `internal: true` network but with the new/fixed companion images, I can see that the DNS-queries to `google` are indeed gone (👍🏻) and -- as a consequence -- the `HTTP(s)` requests, too (👍🏻).

As I already mentioned, I cannot say anything about potential problems with access to `fonts.googleapis.com` being blocked. My "test set" only contains a simple `plantuml mindmap`, a simple `bpmn diagram`, a simple `mermaid gitgraph` and a very simple `excalidraw text box`. Those worked fine.

